PR #9, opened 2021-07-20
Return options fix

dannedelcu:
Fixed crashes when client makes a return request or the return settings page is accessed.

### DIFF
--- a/react/admin/ReturnsSettings.tsx
+++ b/react/admin/ReturnsSettings.tsx
@@ -168,7 +168,7 @@ class ReturnsSettings extends Component<any, any> {
             paymentCard,
             paymentVoucher,
             payments: updatedPayments,
-            options: json[0].options,
+            options: json[0].options??[],
             excludedCategories: JSON.parse(json[0].excludedCategories),
             loading: false
           });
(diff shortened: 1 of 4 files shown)


react/admin/ReturnsSettings.tsx
```diff
@@ -168,7 +168,7 @@ class ReturnsSettings extends Component<any, any> {
             paymentCard,
             paymentVoucher,
             payments: updatedPayments,
-            options: json[0].options,
+            options: json[0].options??[],
```
mariusfiliutamindcell, line 171: use formatter from project tslint
also null check all possibilities
Ex.:
 ```javascript
json?.[0]?.options ?? []
```

File: react/components/RequestForm.tsx [hunk 2]
```diff
@@ -230,7 +230,7 @@ class RequestForm extends Component<Props> {
       }
     ];
 
-    if (options.length !== 0) {
+    if (options && options.length !== 0) {
```
mariusfiliutamindcell, line 233: U can use Optional Chaining here aswell ( `?.` operator )
Ex.:

```javascript
if (options?.length !== 0) {
```

PR #11, opened 2021-07-22
Add return labels and credit card returns

tsheng1:
### Return Labels
- Added return label functionality through integration with Easypost
- The Easypost app (found at [https://github.com/vtex-apps/easypost](url)) must be installed in the workspace. Make sure to fill out the client key and return address fields in the Admin page

### Credit Card Refunds
- Added ability to process returns from orders using a credit card as payment
- Refund invoice will be automatically generated when the status of a refund request is set to `refunded`


ebigler:
Can someone please take care of this code review for us?

PR #12, opened 2021-07-29
Exposes api endpoint to elefantdev.rma-proxy app to allow rma creation

dannedelcu:


duxy (commented):
This need to be a fork of this project and whitelisted for elefant, because this does not apply to other customers.

PR #19, opened 2021-09-09
Add Tax Calculations to Returns

tsheng1:
### Added
- Add tax logic to credit card returns
### Removed
- Removed label creation button from customer side


react/admin/ReturnForm.tsx
```diff
@@ -300,11 +301,73 @@ class ReturnForm extends Component<any, any> {
...
+    if (request) {
+      try {
+        await fetch(`${fetchPath.getOrder}${request.orderId}`)
+        .then((response) => response.json())
+        .then((res) => {
+          taxItems = res.items
+          return Promise.resolve(res)
+        })
+      } catch (e) {
+        // console.log(e)
```
ancasolot, line 327: Maybe we should set an error message here

react/admin/ReturnForm.tsx
```diff
@@ -325,6 +388,10 @@ class ReturnForm extends Component<any, any> {
       if (statusInput !== request.status) {
         requestData = { ...requestData, status: statusInput }
 
+        if (statusInput === requestsStatuses.refunded) {
```
ancasolot, line 391: What is this if doing?

ancasolot (approved):

PR #36, opened 2021-11-02
[RMA-19] Determine if an order is partially refuned

Arturo777:
- When a request was made from the OMS the available quantity on the available orders was not updated.

- When the user was using the verify package API to set the restock fee and shipping value to refund, the store front during the pending verification step, showed a 0 value on the refunded shipping value.


tsheng1 (approved):

albertm805 (approved):
LGTM

PR #40, opened 2021-11-25
Fixes

khrizzcristian:
- Access for telesales user
- GetProfile request from FE switched back to fetch instead of axios. 
- Quantity and availability of a product if there is already a return request placed for that product.


andreeacalinvtex (approved):
The code seems to fix the presented issues.

robertdiaconeasa (approved):

PR #46, opened 2021-12-15
Updates 

khrizzcristian:
- The card refund method is valid only if the customer has paid for the order with the card.
- Fix shipping value available in frontend
- Fix shipping value refunded in frontend
- Email fixes
- The total value of the products was multiplied by 100 in frontend
- The value of the product available for refund is displayed as "NAN" if the product did not have the tax set
- "Submitted by" was blank in history timeline because of a prop.
- Check to prevent users from viewing other users' return requests.


andreeacalinvtex (approved):

tibiiordachevtex (approved):

albertm805 (approved):
LGTM

PR #51, opened 2022-01-31
Feat: add other option checkbox in admin.

asemqaffaf:

In admin section where users can set the "Return Reasons"
this list is dynamic and it renders when the customer wants to create a new return on the front end.
However in the Return Reasons section there is no "Other" option that the user can turn it on or off so that is will be shown on the front end.
The frontend has (Other) option as a default return reason.
so we add a new checkbox on the admin to activate or deactivate the other return reason, so you can find Enable Other Option in this [link](https://newreturnfixes--yamamayqa.myvtex.com/admin/returns/settings)

**In the Web side, you have to have an account in yamamayqa to test the feature.** 

Go to the store, create an Account or you can use an exiting account. Go browse the catalog, add something to your cart, place the order via the checkout, wait a couple of minutes, and then access the store’s My Account page, there you will see that you’ll have a list of available orders to be returned, then you can select the item that you want to return, select the item that you want to return, then you will find a dropdown list that you can find inside them the reasons to return the item, if you enable Other Option in the admin vice versa then you can find it in that dropdown list,
note: you can use this [link](https://newreturnfixes--yamamayqa.myvtex.com/account?__bindingAddress=www.yamamayqa.com/it_it#/my-returns/add) here to see your returns items.

#### How to test it?

[Workspace](https://newreturnfixes--yamamayqa.myvtex.com/)

Screen Shot 2022-01-31 at 6 27 18 PM

[Admin URL](https://newreturnfixes--yamamayqa.myvtex.com/admin/returns/settings)

Screen Shot 2022-01-31 at 6 29 07 PM

[Store URL](https://newreturnfixes--yamamayqa.myvtex.com/account?__bindingAddress=www.yamamayqa.com/it_it#/my-returns/add)

#### How does this PR make you feel? [:link:](http://giphy.com/)

this-could-work-possibility

### DIFF
--- a/messages/context.json
+++ b/messages/context.json
@@ -22,6 +22,8 @@
   "settings.addCustomOption_Name": "Return option",
   "settings.addCustomOption_Days": "Maximum return days",
   "settings.addCustomOption_Submit": "ADD OPTION",
+  "settings.otherOptions": "Other Options",
+  "settings.otherOptions_enable": "Enable Other Options",
   "returns.orderId": "Order ID",
   "returns.requestId": "Request ID",
   "returns.sequenceNumber": "Sequence No",
--- a/messages/context.json
+++ b/messages/context.json
@@ -23,7 +23,7 @@
   "settings.addCustomOption_Days": "Maximum return days",
   "settings.addCustomOption_Submit": "ADD OPTION",
   "settings.otherOptions": "Other Options",
-  "settings.enableOtherOption": "Enable Other Options",
+  "settings.otherOptions_enable": "Enable Other Option",
   "returns.orderId": "Order ID",
   "returns.requestId": "Request ID",
   "returns.sequenceNumber": "Sequence No",
--- a/messages/context.json
+++ b/messages/context.json
@@ -23,7 +23,7 @@
   "settings.addCustomOption_Days": "Maximum return days",
   "settings.addCustomOption_Submit": "ADD OPTION",
   "settings.otherOptions": "Other Options",
-  "settings.otherOptions_enable": "Enable Other Options",
+  "settings.enableOtherOption": "Enable Other Options",
   "returns.orderId": "Order ID",
   "returns.requestId": "Request ID",
   "returns.sequenceNumber": "Sequence No",
--- a/messages/context.json
+++ b/messages/context.json
@@ -22,7 +22,7 @@
   "settings.addCustomOption_Name": "Return option",
   "settings.addCustomOption_Days": "Maximum return days",
   "settings.addCustomOption_Submit": "ADD OPTION",
-  "settings.otherOptions": "Other Options",
+  "settings.otherOptions": "Other Option",
   "settings.otherOptions_enable": "Enable Other Option",
   "returns.orderId": "Order ID",
   "returns.requestId": "Request ID",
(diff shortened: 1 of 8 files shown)


asemqaffaf:

@filafb 
@Git-the-Sanz

react/admin/ReturnsSettings.tsx
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/explicit-member-accessibility */
-import React, { Component } from 'react'
+import { Component, Fragment } from 'react'
```
filafb, line 2: Any reason to remove React from context here? I think we need to keep it since we have JSX in the file and we are using React v16. Eslint should got this, but I think there is some miss configuration here.

react/components/RequestForm.tsx
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/explicit-member-accessibility */
-import React, { Component } from 'react'
+import { Component } from 'react'
```
filafb, line 2: Same thing here about having React in the context.

filafb (changes requested):
Hi Asem, thanks for the PR. Please, make sure you have ran `yarn install` in the root folder. I think there are some eslint / prettier errors that would have been caught. The best way here is to reset your commit, run the command, than commit it again.

filafb:
Another thing we should consider in the future, related to this feature, is to make a form validation in the back end. Right now, there is nothing blocking me to make a request using POSTMAN and sending the reason as `other`. In the back end, we should validate that the reason is one of the list set in the settings plus other reason, if the flag is off. 

c.c. @cesarocampov

cesarocampov:
> Another thing we should consider in the future, related to this feature, is to make a form validation in the back end. Right now, there is nothing blocking me to make a request using POSTMAN and sending the reason as `other`. In the back end, we should validate that the reason is one of the list set in the settings plus other reason, if the flag is off.
> 
> c.c. @cesarocampov

@filafb I agree with you. I have added it as a task.

messages/context.json
```diff
@@ -22,6 +22,8 @@
   "settings.addCustomOption_Name": "Return option",
   "settings.addCustomOption_Days": "Maximum return days",
   "settings.addCustomOption_Submit": "ADD OPTION",
+  "settings.otherOptions": "Other Options",
+  "settings.otherOptions_enable": "Enable Other Options",
```
Git-the-Sanz, line 26: A nitpick here, but I would phrase it as:
`'Other' option`
&
`Enable 'Other' option`
Implying that the option for 'Other' is being enabled, and not that there are 'Other options' available.

Git-the-Sanz (changes requested):
Just a `messages` adjustment

messages/context.json
asemqaffaf: actually i was trying to follow the same pattern of the previous settings variables, as you can see in line 24 and 23 they add underscore postfix the name of the variables, but yeah i will change it as what you see it's better.

Git-the-Sanz: Sorry, not the variables itself. But the TEXT

asemqaffaf: ohh i see, i will change it now

filafb (approved):

filafb:
@cesarocampov Ready to merge!

PR #52, opened 2022-02-02
implement a new step called Processing return form.

asemqaffaf:

In admin section where users can view and set the "Requests"
Requests page has list of items that the customers want to return, when select one of these so you will present Request information page that contains all of that information for a particular item that the customer wants to return.
In the Request information page there is a timeline and in the bottom of that timeline there is a dropdown list that can be update from the Store's admin side (base on the progress that happening on that item), so we add a new step to the timeline and dropdown list called `processing` or `under processing`. so this option can update the timeline as well as the client side (store) of that step as a current status of returning the item.

In the Store side (Web) you can see the new step in `My returns` page.
if you want to reach that page you have to have an account and you did already a purchase and return it already. 
so in this page you can see a list of requests that you've made to return. and you can see the status for each one of the item. so the new update should carry `processing` or `under processing` status to that list. and also when you select any item in the list, you will see all the information about that item and the timeline process of that item, there you will find the new timeline step which is called `processing` or `under processing` so this will inform the client that there item is under processing. 

**In the Web side, you have to have an account in yamamayqa to test the feature.** 

**Admin**
Go to Applications, from `Returns` section select `Request` there you will find a list of all items that all the client users wanted to return as well as status for each on of these, select any item, then you can see Request information page, there you can find the timeline and dropdown list that contains `processing` or `under processing` options.
please check out this [link](https://returnapptimeline--yamamayqa.myvtex.com/admin/returns/46359948-8439-11ec-835d-028f2cdfe527/details/).

**Store:**
Go to the store, create an Account or you can use an exiting account. Go browse the catalog, add something to your cart, place the order via the checkout, wait a couple of minutes, and then access the store’s My Account page, there you will see that you’ll have a list of available orders to be returned, then you can select the item that you want to return, select the item that you want to return, then go to `my returns` you will find  list of items that you want to return include  the status of each one of them, select one of these items then you see all in formations about that include the timeline that i've mentioned before,
note: you can use this [link](https://returnapptimeline--yamamayqa.myvtex.com/account?__bindingAddress=www.yamamayqa.com/it_it#/my-returns/details/46359948-8439-11ec-835d-028f2cdfe527) here to see an item with its returning info include the timeline steps.

#### How to test it?

[Workspace](https://returnapptimeline--yamamayqa.myvtex.com/)

**Admin:**
this is the list of items that all users wanted to return
Screen Shot 2022-02-02 at 3 51 31 PM

this is the request information page with `NEW` status (init. status).
Screen Shot 2022-02-02 at 4 35 41 PM

this is the request information page with `processing` or `under processing`  status (the new status).
Screen Shot 2022-02-02 at 3 57 09 PM

**Client:**
this is my return list, with status.
Screen Shot 2022-02-02 at 3 51 56 PM

this is the status filter option. (filter by status) 
Screen Shot 2022-02-02 at 3 53 22 PM

this is My returns page for a particular item.
Screen Shot 2022-02-02 at 3 57 45 PM

#### How does this PR make you feel? [:link:](http://giphy.com/)

refund-tight-shirt


asemqaffaf:

@filafb
@Git-the-Sanz

Git-the-Sanz (approved):
Good to go

react/admin/ReturnForm.tsx
```diff
@@ -736,6 +736,25 @@ class ReturnForm extends Component<any, any> {
     let allowedStatuses: any = [{ label: currentStatus, value: status }]
 
     if (status === requestsStatuses.new) {
+      allowedStatuses.push({
```
filafb, line 739: This looks good, but I think we have an opportunity here to start simplifying the component. How about making a helper function to return `allowedStatuses` in a different file?

react/common/utils.tsx
```diff
@@ -381,6 +386,28 @@ export function prepareHistoryData(comment: any, request: any) {
...
+        <span>
+          <FormattedMessageFixed id={messages.processing.id} />
+        </span>
+      ),
+      step: 1,
+      comments: comment.filter(
+        (item: any) => item.status === requestsStatuses.processing
+      ),
+      active:
+        request.status === requestsStatuses.processing ||
```
filafb, line 401: I think there is some room here to make a reusable function too. For every single step, we are comparing the previous status and I can see it leading to some bugs in the future (e.g. a new status is added and we don't include it in the comparison). Having the logic in a helper function, might be easier to update it. What do you think?

filafb (changes requested):
Hi Asem, all looks good for me. I added two comments that I would love to hear your thoughts. We need to start simplifying this code and I think we have some opportunity in this PR to do so.

Thanks!

react/admin/ReturnForm.tsx
asemqaffaf: yeah i am agree with you, so i will do it

react/common/utils.tsx
asemqaffaf: yeah that's true! i will make a helper function for it

asemqaffaf:
hello @filafb, i've updated the PR, so could you please review the changes? 
please feel free to ask me any questions

PR #53, opened 2022-02-08
Fix: search pagination result

asemqaffaf:

In admin section where users can view and set the "Returns List" table
Returns page has list of items that the customers want to return, when you use the pagination option for example next page and try to search for any item you will get an empty list and then you have to go back to the first page to see the results. so this PR is fixes that issue by navigate you automatically to the first page that contains the search results.  

**Admin**
Go to Applications, from `Returns` section select `Request` there you will find a list of all items that all the client users wanted to return so you can use this [link](https://searchpaginationresult--yamamayqa.myvtex.com/admin/returns/requests) to test the search for example you can use `sequenceNumber` 

#### How to test it?

[Workspace](https://searchpaginationresult--yamamayqa.myvtex.com/)

**Admin:**
this is the list of items that all users wanted to return with applied search by `sequenceNumber`
 image


asemqaffaf:

@filafb
@Git-the-Sanz

filafb (changes requested):
Hi @asemqaffaf! I got some weird behavior filtering stuff:
- It always take me to the second page with label `16 - 30 of 78` when filtering by status. If I try to go to the first page using the arrow, it shows me a empty state - with label as `1 - 15 of`;
- When I filter by request id, it takes to the first page, but the pagination label is `16 - 1 of 1` and the arrow function to see the previous page is enabled. When I click it, it shows me a empty state page.
Do you mind taking a look at those behaviors, please?

react/components/ReturnTableModal.tsx
```diff
@@ -0,0 +1,64 @@
...
+const ReturnTableModal = ({
+  selectedRequestProducts,
+  isModalOpen,
+  handleModalToggle,
+}: ReturnTableModalProps) => {
+  return (
+    <Modal centered isOpen={isModalOpen} onClose={handleModalToggle}>
+      <div className="dark-gray">
+        {selectedRequestProducts.length ? (
+          <table className={`${styles.table} ${styles.tableModal}`}>
```
filafb, line 22: It's better if we use the `useCSSHandle` hook to apply the classes. Doing the way we are doing here, if someone in the future deletes the class in the css file, it will appear as `undefined` in the DOM.

filafb (changes requested):
Great work Asem! Not easy doing a refactor like this. When you have a minute, can you review my comments?

react/common/ReturnsTableSchema.tsx
```diff
@@ -0,0 +1,79 @@
...
+const ReturnsTableSchema = ({
+  navigate,
+  handleViewRequest,
+}: ReturnsTableSchemaProps) => {
+  return {
+    properties: {
+      id: {
+        title: <FormattedMessage id="returns.requestId" />,
+        sortable: true,
+        width: 350,
```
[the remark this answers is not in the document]
asemqaffaf: i've checked `Table styleguide` documentation, but i could't find that we can make overflow to a certain column

react/admin/ReturnsTable/ReturnsTableContent.tsx
```diff
@@ -80,88 +77,93 @@ class ReturnsTableContent extends Component<IProps, IState> {
...
+        this.setState({ error: returns.error, tableIsLoading: false })
+      } else {
+        this.setState({
+          itmes: returns.data,
+          totalItems: returns.pagination?.total,
+          tableIsLoading: false,
+        })
+      }
+    } catch (error) {
+      this.setState({ error: error.message })
```
filafb, line 145: Here I think we should set state to `tableIsLoading: false`

filafb (changes requested):

react/admin/ReturnsTable/ReturnsTableContent.tsx
asemqaffaf: correct! i think there are two places to add `tableIsLoading: false`

filafb: You can always add a finally block, like [here](https://github.com/vtex-apps/admin-pages/blob/bdf655a3530b528292af89c1ee7f9b4627572c8f/react/components/admin/store/PWAForm/index.tsx#L119-L121)

filafb (approved):
LGTM 🚀 

@cesarocampov Ready to go, with the feature to make the `sequenceNumber` column not sortable.

PR #56, opened 2022-02-16
feat: create func to fetch near pickuppoints by geoCoordinates

zekivtex:
What this PR changes:
Backend:

Frontend and admin:
Functionality to add near pickup points as a delivery option to the user. The default behaviour is to not show this options, in order to display them we have a new settings on the return settings app (admin).

How to test it:
[this workspace](https://zeki--vtexspain.myvtex.com/account#/my-returns).
On this workspace create a return request and click on the toggle aside the default delivery option. Select a pickup point, go to the next step and the preview you can see that the address is updated with the pickup point data.


react/package.json
```diff
@@ -32,19 +32,19 @@
...
     "vtex.my-account-commons": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.my-account-commons@1.6.0/public/@types/vtex.my-account-commons",
     "vtex.react-portal": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.react-portal@0.4.1/public/@types/vtex.react-portal",
     "vtex.render-runtime": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.render-runtime@8.132.3/public/@types/vtex.render-runtime",
-    "vtex.session-client": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.session-client@1.0.2/public/@types/vtex.session-client",
-    "vtex.store": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.store@2.120.1/public/@types/vtex.store",
-    "vtex.store-components": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.store-components@3.154.0/public/@types/vtex.store-components",
-    "vtex.store-graphql": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.store-graphql@2.146.2/public/@types/vtex.store-graphql",
+    "vtex.return-app": "https://pickup--powerplanet.myvtex.com/_v/private/typings/linked/v1/vtex.return-app@2.17.0+build1644963541/public/@types/vtex.return-app",
```
filafb, line 39: Please, remove this line. When you run `vtex setup --typings` it gets added here, but since the reference is a linked app, it breaks when you try to use this branch anywhere else.

node/package.json
```diff
@@ -17,19 +17,19 @@
...
     "vtex.my-account-commons": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.my-account-commons@1.6.0/public/@types/vtex.my-account-commons",
     "vtex.react-portal": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.react-portal@0.4.1/public/@types/vtex.react-portal",
     "vtex.render-runtime": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.render-runtime@8.132.3/public/@types/vtex.render-runtime",
-    "vtex.session-client": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.session-client@1.0.2/public/@types/vtex.session-client",
-    "vtex.store": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.store@2.120.1/public/@types/vtex.store",
-    "vtex.store-components": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.store-components@3.154.0/public/@types/vtex.store-components",
-    "vtex.store-graphql": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.store-graphql@2.146.2/public/@types/vtex.store-graphql",
+    "vtex.return-app": "https://pickup--powerplanet.myvtex.com/_v/private/typings/linked/v1/vtex.return-app@2.17.0+build1644963541/public/@types/vtex.return-app",
```
filafb, line 24: Please, remove this line. When you run vtex setup --typings it gets added here, but since the reference is a linked app, it breaks when you try to use this branch anywhere else.

node/utils/utils.ts
```diff
@@ -1,3 +1,33 @@
...
+    case 403: {
+      throw new ForbiddenError(e)
+    }
+
+    case 400: {
+      throw new UserInputError(e)
+    }
+
+    default:
+      break
```
filafb, line 27: I think it's better to throw the original error here. Otherwise, any error not listed here will not be thrown.

node/resolvers/checkout.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  try {
+    const closestPickupPointsData = await checkout.getNearestPickupPoints(
+      lat,
+      long
+    )
+
+    return closestPickupPointsData
+  } catch (error) {
+    console.log(error.response, '----specific errror----')
```
filafb, line 18: We should remove the try / catch here. The way it's now, if something goes wrong, we will not pass the error to graphql.

filafb (changes requested):
Nice work @zekivtex! Thanks! I added some comments to it. Please let me know if you have any question.

Also, how are we planning to capture the lat and long from the user?

zekivtex:
> Nice work @zekivtex! Thanks! I added some comments to it. Please let me know if you have any question.
> 
> Also, how are we planning to capture the lat and long from the user?

Thank you Fila!

react/components/RequestForm.tsx
```diff
@@ -706,4 +815,23 @@ class RequestForm extends Component<Props> {
...
+  graphql(NEAR_PICKUP_POINTS, {
+    options: () => ({
+      variables: {
+        lat,
+        long,
+      },
+      ssr: false,
+    }),
+    name: 'NearestPickupPoints',
+  })
```
filafb, line 836: Instead of wrapping all in a HOC, let's start migrating to a function component. Please create a function component that wraps this one and pass down the props needed - the query and intl. We are doing something like that [here](https://github.com/vtex-apps/return-app/blob/aeaec1a9892a9bf63a12d61d2ae4c15cd316073d/react/StoreMyReturnsPageAdd.tsx#L24).

filafb (changes requested):
Hey @zekivtex. I am trying to test the front end, but I cannot see the new feature and the admin doesn't seem to update the schemas. It gets stuck in the spinner. Can you check it please?
Screen Shot 2022-02-24 at 10 35 24 AM
Screen Shot 2022-02-24 at 10 35 16 AM

PR #58, opened 2022-02-17
Update schema

filafb:
This a draft PR so we can discuss the approach to update schemas.

### Current Problem
For every change in the schema, a store admin needs to go to the settings page in order to trigger the schema update. If they don't that, it can lead to problems when saving and searching documents.

The current process is the following:
1. The app checks the schema when the component for the Admin Settings renders [here](https://github.com/vtex-apps/return-app/blob/1bef8ee0be84fcf14b29f81df5a4f1a435be05a3/react/admin/ReturnsSettings.tsx#L206);
2. The browser fetches all the schemas, and compare every one with its pair in the code [here](https://github.com/vtex-apps/return-app/blob/1bef8ee0be84fcf14b29f81df5a4f1a435be05a3/react/common/utils.tsx#L582);
3. Then, the browsers call the back end, that will update all the schemas [here](https://github.com/vtex-apps/return-app/blob/1bef8ee0be/node/clients/masterdata.ts#L57-L124).

### My suggestion
Devise a logic all in the back end to trigger the schema update and place it in key processes, like when __store users are creating new orders__ AND the __admin visits the settings page__. 

The new function on `node/utils/checkAndUpdateSchemas.ts` can be called inside any resolver and handles all the logic.

The decision process is the following:

[](https://mermaid-js.github.io/mermaid-live-editor/edit#pako:eNpVkV9rgzAUxb_KJU8rtF_Ah5VNrXWDvnQrjOhDMNcaqokkcVDU7774j6nkQfI75-TkpiWZ4kg8ctesLuArSCS4762NDbC6hlLIB_JjP2_D4fAK3Q-aDnx6URa4kphu4EV1EFC_wOwBt3dmFhyMOGxvqI1QEpqaM_sfHa6iTy-Bi91twBDr0-_RBL9zhlvrI_xRGtEILZiswIqZmUQjOVNfVTXTCNcNPTsKJ6UBXfBzto6OeJjDJAYu8hw1SrtUjleVP5ZqkzjdKIbun3S4U0r2pEJdMcHdzNtBlRDrHJgQz_1yzFlT2oQksnfSaUYhF1Zp4uWsNLgnrLHq-pQZ8axucBEFgrknrGZV_wchQJQq)

### How to test the function

In this [workspace](https://fila--vtexspain.myvtex.com/_v/private/vtex.return-app@2.17.0/graphiql/v1?query=mutation%20%7B%0A%20%20createOrUpdateSchemas%0A%7D), one can trigger the function.

To see the effects, update any of the schema using Postman. Run the mutation, then check the schema on postman. It should be updated back to the original version.

Update the schema - missing field `skuId` inside `v-indexed`:
```curl
curl --location --request PUT 'https://vtexspain.vtexcommercestable.com.br/api/dataentities/ReturnApp/schemas/returnProducts' \
--header 'VtexIdclientAutCookie: {{authToken}}' \
--header 'Content-Type: application/json' \
--data-raw '{
    "properties": {
        "refundId": {
            "type": "string",
            "IsRelationship": true
        },
        "orderId": {
            "type": "string",
            "IsRelationship": true
        },
        "userId": {
            "type": "string",
            "IsRelationship": true
        },
        "imageUrl": {
            "type": "string"
        },
        "skuId": {
            "type": "string"
        },
        "sku": {
            "type": "string"
        },
        "productId": {
            "type": "string"
        },
        "ean": {
            "type": "string"
        },
        "brandId": {
            "type": "string"
        },
        "brandName": {
            "type": "string"
        },
        "skuName": {
            "type": "string"
        },
        "manufacturerCode": {
            "type": "string"
        },
        "unitPrice": {
            "type": "integer"
        },
        "quantity": {
            "type": "integer"
        },
        "totalPrice": {
            "type": "integer"
        },
        "goodProducts": {
            "type": "integer"
        },
        "reasonCode": {
            "type": "string"
        },
        "reason": {
            "type": "string"
        },
        "condition": {
            "type": "string"
        },
        "status": {
            "type": "string"
        },
        "dateSubmitted": {
            "type": "string",
            "format": "date-time"
        },
        "type": {
            "type": "string"
        }
    },
    "v-security": {
        "allowGetAll": true,
        "publicFilter": [
            "refundId",
            "orderId",
            "userId",
            "imageUrl",
            "skuId",
            "sku",
            "productId",
            "ean",
            "brandId",
            "brandName",
            "skuName",
            "manufacturerCode",
            "unitPrice",
            "quantity",
            "totalPrice",
            "goodProducts",
            "reasonCode",
            "reason",
            "condition",
            "status",
            "dateSubmitted",
            "type"
        ],
        "publicJsonSchema": false
    },
    "v-immediate-indexing": true,
    "v-cache": false,
    "v-default-fields": [
        "id",
        "createdIn",
        "refundId",
        "orderId",
        "userId",
        "imageUrl",
        "sku",
        "skuId",
        "productId",
        "ean",
        "brandId",
        "brandName",
        "skuName",
        "manufacturerCode",
        "unitPrice",
        "quantity",
        "totalPrice",
        "goodProducts",
        "reasonCode",
        "reason",
        "condition",
        "status",
        "dateSubmitted",
        "type"
    ],
    "v-indexed": [
        "id",
        "createdIn",
        "refundId",
        "orderId",
        "userId",
        "sku",
        "skuName",
        "status",
        "type"
    ]
}'
```
Get the schema - field `skuId` inside `v-indexed` should be back.

```curl
curl --location --request GET 'https://vtexspain.vtexcommercestable.com.br/api/dataentities/ReturnApp/schemas/returnProducts' \
--header 'VtexIdclientAutCookie: {{authToken}}' 
```

### DIFF
--- a/node/clients/mdFactory.ts
+++ b/node/clients/mdFactory.ts
@@ -86,4 +86,79 @@ export class MDFactory extends MasterData {
       fields: productReturnFields,
     })
   }
+
+  public updateSettingsSchema(schemaBody: typeof SETTINGS_SCHEMA) {
+    return this.createOrUpdateSchema({
+      dataEntity: this.dataEntity,
+      schemaName: this.schemas.settingsSchema.name,
+      schemaBody,
+    })
+  }
+
+  public updateRequestSchema(schemaBody: typeof RETURNS_SCHEMA) {
+    return this.createOrUpdateSchema({
+      dataEntity: this.dataEntity,
+      schemaName: this.schemas.returnSchema.name,
+      schemaBody,
+    })
+  }
+
+  public updateCommentsSchema(schemaBody: typeof COMMENTS_SCHEMA) {
+    return this.createOrUpdateSchema({
+      dataEntity: this.dataEntity,
+      schemaName: this.schemas.commentsSchema.name,
+      schemaBody,
+    })
+  }
+
+  public updateHistorySchema(schemaBody: typeof HISTORY_SCHEMA) {
+    return this.createOrUpdateSchema({
+      dataEntity: this.dataEntity,
+      schemaName: this.schemas.statusHistorySchema.name,
+      schemaBody,
+    })
+  }
+
+  public updatetProductsSchema(schemaBody: typeof PRODUCTS_SCHEMA) {
+    return this.createOrUpdateSchema({
+      dataEntity: this.dataEntity,
+      schemaName: this.schemas.productsSchema.name,
+      schemaBody,
+    })
+  }
+
+  public getSettingsSchema() {
+    return this.getSchema({
+      dataEntity: this.dataEntity,
+      schema: this.schemas.settingsSchema.name,
+    })
+  }
+
+  public getRequestSchema() {
+    return this.getSchema({
+      dataEntity: this.dataEntity,
+      schema: this.schemas.returnSchema.name,
+    })
+  }
+
+  public getCommentsSchema() {
+    return this.getSchema({
+      dataEntity: this.dataEntity,
+      schema: this.schemas.commentsSchema.name,
+    })
+  }
+
+  public getHistorySchema() {
+    return this.getSchema({
+      dataEntity: this.dataEntity,
+      schema: this.schemas.statusHistorySchema.name,
+    })
+  }
+
+  public getProductsSchema() {
+    return this.getSchema({
+      dataEntity: this.dataEntity,
+      schema: this.schemas.productsSchema.name,
+    })
+  }
 }
--- a/node/utils/checkAndUpdateSchemas.ts
+++ b/node/utils/checkAndUpdateSchemas.ts
@@ -30,11 +30,13 @@ const isSchemaUpdated = async (ctx: Context) => {
   return false
 }
 
-export const updateSchema = async (ctx: Context) => {
+export const checkAndUpdateSchemas = async (ctx: Context) => {
   const {
     clients: { vbase, mdFactory },
     vtex: { logger },
   } = ctx
+
+  let hasUpdated = false
 
   try {
     const updated = await isSchemaUpdated(ctx)
@@ -75,24 +77,29 @@ export const updateSchema = async (ctx: Context) => {
     if (!isDeepStrictEqual(settingsSchema, SETTINGS_SCHEMA)) {
       updateSettingsSchemaPromise =
         mdFactory.updateSettingsSchema(SETTINGS_SCHEMA)
+      hasUpdated = true
     }
 
     if (!isDeepStrictEqual(requestSchema, RETURNS_SCHEMA)) {
       updateRequestSchemaPromise = mdFactory.updateRequestSchema(RETURNS_SCHEMA)
+      hasUpdated = true
     }
 
     if (!isDeepStrictEqual(commentsSchema, COMMENTS_SCHEMA)) {
       updateCommentsSchemaPromise =
         mdFactory.updateCommentsSchema(COMMENTS_SCHEMA)
+      hasUpdated = true
     }
 
     if (!isDeepStrictEqual(historySchema, HISTORY_SCHEMA)) {
       updateHistorySchemaPromise = mdFactory.updateHistorySchema(HISTORY_SCHEMA)
+      hasUpdated = true
     }
 
     if (!isDeepStrictEqual(productsSchema, PRODUCTS_SCHEMA)) {
       updateProductsSchemaPromise =
         mdFactory.updatetProductsSchema(PRODUCTS_SCHEMA)
+      hasUpdated = true
     }
 
     await Promise.all([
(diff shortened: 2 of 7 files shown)


zekivtex (approved):

node/utils/checkAndUpdateSchemas.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    await Promise.all([
+      updateSettingsSchemaPromise,
+      updateRequestSchemaPromise,
+      updateCommentsSchemaPromise,
+      updateHistorySchemaPromise,
+      updateProductsSchemaPromise,
+    ])
+  } catch (error) {
+    if (error.response.status !== 304) {
```
Git-the-Sanz, line 113: @filafb @zekivtex @cesarocampov 
Would be there any advantage trying to catch errors from independent promises, to later log or return?

I was thinking like, logging which schema failed and why, and have a Splunk trigger if this is happening in a `master` workspace, as a red flag for us to contact the owners of missuse 🚩

node/clients/mdFactory.ts
```diff
@@ -90,4 +90,79 @@ export class MDFactory extends MasterData {
       fields: productReturnFields,
     })
   }
+
+  public updateSettingsSchema(schemaBody: typeof SETTINGS_SCHEMA) {
```
Git-the-Sanz, line 94: @filafb @zekivtex 
Couldn't this be made only 1 method? So we later recursively check for schemas when accessing `checkAndUpdateSchemas()`

node/utils/checkAndUpdateSchemas.ts
filafb: Correct me if I am wrong, but to catch independent errors, we need to run the calls synchronously. Doing so, the tradeoff can be very expensive. On postman, a call to update a schema can take a couple of seconds. If we make the calls synchronous, that time will pilled up. An alternative, could be to use `Promise.allSettle`, but TS doesn't allow me to use that (and tbh, I am not full aware about its performance).

And in theory, the errors regarding a specific schema should happen during development, since the only missuse I see here is on a dev side.

node/clients/mdFactory.ts
filafb: You mean, one method that we pass the schema as an argument when calling it or a method like `createOrUpdateAllSchemas` and then we do the logic to update all schemas inside it?

node/utils/checkAndUpdateSchemas.ts
Git-the-Sanz: With promise.all the first error would trigger the catch, and stop the rest; I was thinking of logging which one was the problematic one, so If you are developing modifying more than 1 schema, you could see the exact one. But maybe there's no need of knowing that haha

node/clients/mdFactory.ts
Git-the-Sanz: Yeah, I mean, there's only 1 dataEntity and a small enum of the schemas' names. Why we cannot have only 1 public method? And later in `checkAndUpdateSchemas` we can map all the schemas with only 1 function.

filafb: Got it! We def could have only one public method. I was back and forth with that implementation, and tbh, I don't have a strong opinion which one is better.
  
The main reason I decided having different methods were: 
1. We are already creating a dictionary for the schemas inside the `mdFactory` client. Creating the enum, I saw it as a possibly two different source of truth for the schemas;
2. Having different methods allow us to be more strict with the schema type we are passing for every single method. The client manages the other arguments (data entity and schema name) for us;
3. In the next major I think we should go for a different data entity for every entity we have (`settings`, `requests`, `products`, `comments`, `history`). One method per entity mimics how it would be (one client per data entity). This structure would reduce the pressure on the indexation process. Per the [documentation](https://developers.vtex.com/vtex-rest-api/docs/master-data-schema-lifecycle#creating-a-schema), there is a background process validating all documents when creating a schema. The way the app is structured right now, when we update the schema for `products`, all documents, including the ones not related to products, will be considered. 

What do you think? Still consider having only one a better approach? Or should we start preparing for the `#3` I mentioned above and create different clients for different entities?

node/utils/checkAndUpdateSchemas.ts
filafb: But is there a way to do that without running the requests synchronously? I think it would be a nice addition, if it doesn't hurt the performance.

node/clients/mdFactory.ts
Git-the-Sanz: Totally in with the `#3`! 😍

node/utils/checkAndUpdateSchemas.ts
Git-the-Sanz: Yeah! The methods from the mdFactory are not catching errors, if we catch and throw them there, each method can have its own custom error object passed along to the promise.all catcher. There we could see who's error it's when using logger.

Or we could also write a catch, for every one of them independently, when assigning them to the `updateSchemaPromise`. That error object would be passed along also. But that will mean that .all will behave exactly as the sugar coated allSettled and not halt with the first error

PR #59, opened 2022-02-20
New setting for payment methods

Git-the-Sanz:

Account holders want a new setting to prevent customers from selecting a different payment method to which the refund will be emitted.

#### How to test it?

[Workspace](https://paymentselection--vtexspain.myvtex.com/)

Ensure the new setting is toggled on/off and review the `refund request form` in my account.

If new setting is toggled ON
Setting on

Customers will not be able to select a different payment method
Request form active

If new setting is toggled OFF
Setting off

Customers will be able to choose a payment method following the payment method options
Request form inactive


react/store/MyReturnsPageAdd.tsx
```diff
@@ -266,6 +266,12 @@ class MyReturnsPageAdd extends Component<any, State> {
       })
     }
 
+    if (this.state.settings.hidePaymentMethodSelection) {
+      this.setState({
+        paymentMethod: 'Default',
```
Git-the-Sanz, line 271: @cesarocampov @filafb @zekivtex 

All refund forms REQUIRE the payment method before creating a request.
If the setting is turned `ON` there's no input for that, so I chose to pass the value as 'Default' when creating the request.

The order payment method is not directly available:

Payment data

We can parse the exact method chosen for the order, but it would require more time, as we should test to prove we are returning the correct values. I think we can backlog it.

Git-the-Sanz: Resolved: we are passing a translatable message.
`Refund with order payment method`

cesarocampov (approved):

zekivtex (approved):

PR #63, opened 2022-02-24
Fix: Show total amount on order details page for admin

filafb:
This is a quick (and **ugly**) fix to the problem preventing admin users to see the total amount of a RMA. The logic calculating the total amount is inside the `handleTaxes` method [here](https://github.com/vtex-apps/return-app/blob/aeaec1a9892a9bf63a12d61d2ae4c15cd316073d/react/admin/ReturnForm.tsx#L327-L402).

To make it more robust, it will be necessary to test different cases where the products have different taxes.

This [workspace](https://rmafixtotalamount--yamamayqa.myvtex.com/admin/returns/8a835cb2-87f8-11ec-835d-029fc0b7bac3/details/) has the code linked.

To compare, this is the [original version](https://yamamayqa.myvtex.com/admin/app/returns/8a835cb2-87f8-11ec-835d-029fc0b7bac3/details/)


zekivtex:
@filafb  @cesarocampov An example with multiple products. 
PD: I thought I have to review this PR too.

filafb:
@zekivtex Thanks for doing so! Did you see any problem in the order you sent? It worked fine here. 
Screen Shot 2022-02-25 at 9 44 17 AM

PR #66, opened 2022-03-02
feat: add pickup points as as a delivery option

zekivtex:
What this PR changes:
Backend:

Frontend and admin:
Functionality to add near pickup points as a delivery option to the user. The default behaviour is to not show this options, in order to display them we have a new settings on the return settings app (admin).

How to test it:
[this workspace](https://pickuppoints--vtexspain.myvtex.com/account#/my-returns).
On this workspace create a return request and click on the toggle aside the default delivery option. Select a pickup point, go to the next step and the preview you can see that the address is updated with the pickup point data.


node/resolvers/checkout.ts
```diff
@@ -0,0 +1,16 @@
+export const nearestPickupPoints = async (
+  _: any,
+  { lat, long }: CheckoutPickupPointsArgs,
+  ctx: Context
+) => {
```
Git-the-Sanz, line 5: It would be great to type what we are returning here!
Something like [this line](https://github.com/vtex-apps/cross-device-cart/blob/V1/node/resolvers/replaceCart.ts#L15)

react/components/RequestForm.tsx
```diff
@@ -104,14 +116,75 @@ const messages = defineMessages({
...
+    this.state = {
+      isPickupPointsSelected: false,
+      selectedPickupPoint: '',
+      dropdownOptionsPickupPoints: [],
+    }
+  }
+
   componentDidMount(): void {
     typeof window !== 'undefined' && window.scrollTo(0, 0)
+    console.log(this.props.settings.enablePickupPoints, 'piopssss')
```
Git-the-Sanz, line 142: 🤭

react/components/RequestForm.tsx
```diff
@@ -104,14 +116,75 @@ const messages = defineMessages({
...
+    lat =
+      this.props.selectedOrder.shippingData.address.geoCoordinates[0].toString()
+    long =
+      this.props.selectedOrder.shippingData.address.geoCoordinates[1].toString()
+  }
```
Git-the-Sanz, line 147: You can improve legibility of these assignments by restructuring the array.

```javascript
const address = {
geoCoordinates: ['asd', 'wasd']
}

let lat
let long

[lat, long] = address.geoCoordinates

console.log(lat) // asd
console.log(long) // wasd
```

node/clients/checkout.ts
```diff
@@ -0,0 +1,44 @@
...
+      },
+    })
+  }
+
+  public getNearestPickupPoints = (lat: string, long: string) => {
+    return this.get<CheckoutOutput>(
+      this.routes.getNearestPickupPoints(lat, long),
+      {
+        metric: 'checkout-nearestPickupPoints',
+      }
```
Git-the-Sanz, line 24: Maybe we could use `.then()` to directly return the `items` array, like we do [here](https://github.com/vtex-apps/cross-device-cart/blob/V1/node/clients/checkout/index.ts#L72)

react/components/RequestForm.tsx
```diff
@@ -734,4 +850,23 @@ class RequestForm extends Component<Props> {
...
+export default compose(
+  graphql(NEAR_PICKUP_POINTS, {
+    options: () => ({
+      variables: {
+        lat,
+        long,
+      },
+      ssr: false,
+    }),
+    name: 'NearestPickupPoints',
```
Git-the-Sanz, line 870: What do you think of changing this name?
Or maybe adding an `alias` to the query on the .gql file, so we don't repeat twice `NearestPickupPoints`

Git-the-Sanz (changes requested):
Good job! 🎉
I will nitpick on some changes tho, to improve legibility.

react/components/RequestForm.tsx
edyespinal: Piopssss!

react/admin/ReturnsSettings.tsx
```diff
@@ -171,6 +175,8 @@ class ReturnsSettings extends Component<any, any> {
             checked: paymentVoucher,
           },
         }
+
+        console.log(json[0], 'optionsss')
```
edyespinal, line 179: 🤭

node/clients/checkout.ts
```diff
@@ -0,0 +1,44 @@
...
+    )
+  }
+
+  protected get = async <T>(url: string, config: RequestConfig = {}) => {
+    try {
+      return await this.http.get<T>(url, config)
```
edyespinal, line 30: The await is added because without it, the catch block would never be reached.

filafb (changes requested):
Two changes that I think it's worth testing / changing. 

Also we should make a component to handle the Pickup point options - especially in the RequestForm component. This would help us migrating this component to `v3`. I don't think this is mandatory for this PR, but it's something to keep in mind for everything we do from now on in this project.

filafb:
Hey @zekivtex! I linked the branch [here](https://pickuppoints--vtexspain.myvtex.com/) and tried to test it, but I am getting an error. Can you check it, please?
Recording 2022-03-18 at 10 34 17

filafb:
Also, when we publish this version, it will break the indexation for the other schemas, preventing users to see the requests the have made and the products associated to it. I think a ugly but effective work around here would be adding all fields we have in the other schemas in the settings (both in the properties and index). This way, the indexation process would not remove the other documents when this schema gets updated. What do you think?

zekivtex:
> Also, when we publish this version, it will break the indexation for the other schemas, preventing users to see the requests the have made and the products associated to it. I think a ugly but effective work around here would be adding all fields we have in the other schemas in the settings (both in the properties and index). This way, the indexation process would not remove the other documents when this schema gets updated. What do you think?

Great idea Fila. As you said I will merged them on another PR.

filafb (changes requested):
Still two things about the UI:

1. The toggle to use pickup points needs to have a label, otherwise the user will not know what it refers to. And it has to be something the store user easily understands. Maybe a tool tip can be helpful here. 
2. When I toggle to use pickup points, on, off and on, the state in the address is wrong / out of sync. (see image: the drop down says a pickup point, but the address details are mine). Maybe we need to clean the state `this.state.selectedPickupPoint` when I visit back my address;
3. There is a `null` value in the address for pickup points. When building the string, we should omit the falsy values.

_Mismatch between drop down value and address_
Screen Shot 2022-03-30 at 10 47 03 AM

_`null` value in the address_
Screen Shot 2022-03-30 at 10 35 33 AM

PR #67, opened 2022-03-03
Handle errors

zekivtex:
What this PR changes:
Improve error handling and use splunk logger method in order to see on the platform.

How to test it:
[this workspace](https://zekiqa--vtexspain.myvtex.com/account#/my-returns).
Add a `throw new Error('any message')` on any defined route go to to the given workspace reload the page and then you can see the error on splunk.
`receiveDocuments.ts` is a good file to test because it loads on the root of the app.

### DIFF
--- a/node/middlewares/getSkuById.ts
+++ b/node/middlewares/getSkuById.ts
@@ -1,13 +1,30 @@
 export async function getSkuById(ctx: Context, next: () => Promise<any>) {
   const {
     clients: { returnApp: returnAppClient },
+    vtex: { logger },
   } = ctx
 
   const { id } = ctx.vtex.route.params
-  const response = await returnAppClient.getSkuById(ctx, id)
 
-  ctx.status = 200
-  ctx.body = response
+  try {
+    const response = await returnAppClient.getSkuById(ctx, id)
+
+    logger.info({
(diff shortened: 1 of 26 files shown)


node/middlewares/getSkuById.ts
```diff
@@ -1,13 +1,34 @@
...
+
+    logger.info({
+      message: 'Get SKU by id successfully',
+      data: response,
+    })
+
+    ctx.status = 200
+    ctx.body = response
+  } catch (e) {
+    logger.error({
```
Git-the-Sanz, line 24: Wouldn't having this catcher here defeat the purpose of the `errorHandler` wrapper?
I mean, the `throw` in line 13 would be catched by this try/catch, rendering the errorHandler catcher redundant.

Git-the-Sanz (changes requested):
Looks good! I would only change how we handle the error in `getSkuById`

node/middlewares/getSkuById.ts
zekivtex: > Looks good! I would only change how we handle the error in `getSkuById`
 jaja yes! I forgot to update that one

edyespinal (approved):

Git-the-Sanz (approved):

PR #71, opened 2022-03-15
Calculate proportional shipping value

zekivtex:
What this PR changes:
Add the option to calculate the delivery value proportionally to the total cost of the returned product(s) in the return request.

How to test it:
Toggle the Shipping value proportional to the total of the products to test the new feature here => [this workspace](https://zekiqa--vtexspain.myvtex.com/admin/returns/settings)
Then go to https://zekiqa--vtexspain.myvtex.com/account#/my-returns and create a return request. When finished go to details and you will see that the shipping value is proportional to the value of the products and the total of the order.

Here is an example with multiple products: https://zekiqa--vtexspain.myvtex.com/admin/returns/c7fe72a1-af9b-11ec-835d-123072a6b6a3/details/
And this one with more than one of the same product: https://zekiqa--vtexspain.myvtex.com/admin/returns/f12f40dc-aed7-11ec-835d-0aaa62e90a93/details/


filafb:
In the UI, I think we should display the shipping value that can be refunded, not the total. In the example below, the value that can be refunded is `2.46` but the UI shows `2.99`. It can be confusing for the admin users. @cesarocampov what is your take here?

Screen Shot 2022-03-30 at 11 49 32 AM
Screen Shot 2022-03-30 at 11 49 44 AM

filafb:
Another UX we need to think about. The value to be refunded in a RMA should be bound to the rule that was active when it was created, and not calculated when a user (admin or store) visits it. Imagine a RMA is created with the flag to proportional is off and all the process to refund happens to the end. If someone goes to the admin and changes the rule, it will affect all the old ones. I think the best approach here is to add a key in the schema to fix the shipping to be refund in the whole lifecycle of the RMA. Any thoughts here? @zekivtex @cesarocampov

zekivtex:
> Another UX we need to think about. The value to be refunded in a RMA should be bound to the rule that was active when it was created, and not calculated when a user (admin or store) visits it. Imagine a RMA is created with the flag to proportional is off and all the process to refund happens to the end. If someone goes to the admin and changes the rule, it will affect all the old ones. I think the best approach here is to add a key in the schema to fix the shipping to be refund in the whole lifecycle of the RMA. Any thoughts here? @zekivtex @cesarocampov

You are right! And yes, I think adding the property it's the best way to do it.

zekivtex:
> Another UX we need to think about. The value to be refunded in a RMA should be bound to the rule that was active when it was created, and not calculated when a user (admin or store) visits it. Imagine a RMA is created with the flag to proportional is off and all the process to refund happens to the end. If someone goes to the admin and changes the rule, it will affect all the old ones. I think the best approach here is to add a key in the schema to fix the shipping to be refund in the whole lifecycle of the RMA. Any thoughts here? @zekivtex @cesarocampov

You are right! And yes, I think adding the property it's the best way to do it.

PR #77, opened 2022-03-29
Admin settings

filafb:
_This PR got way longer than it should be, sorry. When I realized a better to make smaller ones, it was already too late._

This PR brings a new Admin Settings for return app. It's a whole refactor of the component, moving from having a single component handling data fetch, state, and presentational aspects to several smaller components. There are also some changes in the back end and it starts to make the messages in a more standard format.

##### Changes in the back end:

1. A new resolver `categoryTreeName` to get the category tree adjusted to show the whole category tree in its name. The logic to build the name is happening inside the resolver `node/resolvers/categoryTreeName.ts`. Before, this logic was being applied in the browser. It also brings a new client `catalog`;

2. Renamed some fields on ReturnAppSettings graphql types and are some validations happening in the resolver to save new settings here `node/resolvers/appSettings.ts`. The directive handling the validation on translations input field got removed since it was giving errors when the app was installed.

##### Front end:

1. There are two new Providers: Alert and Settings. The first one handles feed to the user when they are doing some operations and can be reused in the other section of the admin (that is why they are in different folders). The second one 
handles all the logic related to settings state, passing data and actions down directly to the components using context. The state is managed using a reducer, so every different component can dispatch data that is under its responsibility. Both providers pass data in a `hook` way.

The main component `RMASettings` renders single components for every section in the settings:

1. Required Options - Handles maxDays and terms and condition fields;
2. Excluded categories - Fetches the category tree data and handles the logic to add and delete categories from the app state;
3. Payment Options - Handles the boolean to allow users to select a different payment method and the checkbox for different payment types;
4. CustomReasons - The most complicated component, since it handles add, edit, delete and translations for different reasons. I broke it in several other components, trying to make them responsible for smaller tasks. Translations for reasons is a new feature. The hierarchy is the following:
```
<CustomReasons> -> Acts as a centralized state to communicate between the other components;
  <CustomReasonModal> -> Handles add and edit custom reasons; 
  <ModalDialog> -> Handles the delete operation. When the delete button is selected, the index of the reason is placed in the component state;
  <TranslationModal> -> Handles the translations.
```

6. GeneralOptions - Handles the other boolean options.

Important to notice that every component commits its own state to the app state. The settings only get submitted when user press the submit button.

#### How to test it?

The code is linked in [this workspace](https://rmav3admin--powerplanet.myvtex.com/admin/returns/settings). To test it, modify the values and save it. After reloading the page, the values saved should populate the fields.


zekivtex (approved):
@filafb Amazing!

PR #82, opened 2022-04-20
V3 return request schema

filafb:
This PR adds the [Master data builder](https://github.com/vtex/builder-hub/tree/master/node/build/masterData), which will trigger the schema creation upon installation. It creates a data entity named `vtex_return_app_returnRequest` and the schema name will be the app version. In a workspace, it will add `-{workspaceName}` to it.

The main idea was to keep the fields presented on the old schemas, to make easy backward compatibility, grouping then in groups so it's easy to manipulate them when needed.

I added some comments to it. Feel free to add more, suggesting changes, better names, etc.


masterdata/returnRequest/schema.json
```diff
@@ -0,0 +1,211 @@
...
+    },
+    "totalReturnAmount": {
+      "type": "integer",
+      "required": true
+    },
+    "sequenceNumber": {
+      "type": "string",
+      "required": true
+    },
+    "status": { "$ref": "#/$defs/status", "required": true },
```
filafb, line 15: The `status` field is here to make it easy to SEARCH for order status. There is an array `refundStatusData` in the schema that will hold all the changes in status the RMA went through. The `status` field has to be sync'ed with it to represent always the latest one in the array. It will be useful having a validation in the process to update the status (a request cannot be processed if the status sent is previous to the current one).

masterdata/returnRequest/schema.json
```diff
@@ -0,0 +1,211 @@
...
+            "required": true,
+            "enum": ["items", "shipping", "tax"]
+          },
+          "value": {
+            "type": "integer",
+            "required": true
+          }
+        }
+      }
+    },
```
filafb, line 33: The `returnTotals` field is an array with the breakdown of the `totalReturnAmount` field. For the `shipping` item, we can store the proportional value if the store has the flag on.

Still not sure we need a type for `tax`. Should it be included in the `items` value?

masterdata/returnRequest/schema.json
```diff
@@ -0,0 +1,211 @@
...
+        "name": {
+          "type": "string"
+        },
+        "email": {
+          "type": "string",
+          "format": "email"
+        },
+        "phoneNumber": { "type": "string", "maxLength": 50 }
+      }
+    },
```
filafb, line 51: The `customerProfileData` groups the user info data. We could store only the `userId` and fetch user info from CL (to be PII complaint) every time we need it, but than we should handle:
1. User will not be able to change name and phone (as it can do it in the current version);
2. It wouldn't be fully backward compatible, since we would lose the information for the cases that the user has changed the name and phone number;
3. How to handle Marketplace and Seller? Does the info related to customer lives on Marketplace or seller? Does the Markeplace has full access to CL from Seller?

masterdata/returnRequest/schema.json
```diff
@@ -0,0 +1,211 @@
...
+        "addressType": {
+          "type": "string",
+          "required": true,
+          "enum": ["pickup-point", "customer-address"]
+        },
+        "returnLabel": {
+          "type": "string"
+        }
+      }
+    },
```
filafb, line 88: The `pickupReturnData` field groups information about where the items will be picked up. It respects the same format in the current schema (e.g. address is built using street name and number), but changes `locality` to `city` to be more aligned with the information in the Order.

It also adds a address type, which can be a `customer-address` or a `pickup-point`.

Here we can discuss the same thing about being PII compliant - Should we only store `addressId` and fetch it from AD?

masterdata/returnRequest/schema.json
```diff
@@ -0,0 +1,211 @@
...
+          "type": "string"
+        },
+        "accountHolderName": {
+          "type": "string"
+        },
+        "paidWithCc": {
+          "type": "boolean"
+        }
+      }
+    },
```
filafb, line 108: The field `refundPaymentData` groups the information about payment refund. The types matches the enum on graphql and has to be modified if we changed it there. 

I added a called paidWithCC, so we can control if we should display or not the credit card option in the UX.

And we should be aware that storing the `iban` might not be PII compliant, but I don't see a work around here.

masterdata/returnRequest/schema.json
```diff
@@ -0,0 +1,211 @@
...
+            "enum": [
+              "newWithBox",
+              "newWithoutBox",
+              "usedWithBox",
+              "usedWithoutBox"
+            ]
+          }
+        }
+      }
+    },
```
filafb: For `items`, I removed all the product information (like image url, skuId, unit price, etc), since we can get it from the directly from the order. To match that, I added the `orderItemIndex`, to keep track of the product inside `order.items`. It might be useful for split orders, or when a product is given for free when buying two units for that, for example.

`verifiedItems` will be filled after the store has verified and approved the return for that item. I am still not convinced we need the field here, since we can get that information from `refundData.items` property (we just need to check if the verify process and the invoice process happens at the same time). We may delete it in the future.

masterdata/returnRequest/schema.json
```diff
@@ -0,0 +1,211 @@
...
+            "type": "object",
+            "properties": {
+              "orderItemIndex": { "type": "integer", "required": true },
+              "quantity": { "type": "integer", "required": true },
+              "restockFee": { "type": "integer" }
+            }
+          }
+        }
+      }
+    },
```
filafb, line 167: `refundData` will be populate when the process is done - invoice to return items is completed.

masterdata/returnRequest/schema.json
```diff
@@ -0,0 +1,211 @@
...
+  },
+  "v-cache": false,
+  "v-default-fields": [
+    "id",
+    "createdIn",
+    "orderId",
+    "sequenceNumber",
+    "dateSubmitted"
+  ],
+  "v-indexed": ["id", "createdIn", "orderId", "customerProfileData"],
```
filafb: For now, only indexing the fields I see we can use to fetch the document. We should keep it always as minimum as possible and NEVER delete a property from it to avoid breaking indexation.

masterdata/returnRequest/schema.json
```diff
@@ -0,0 +1,211 @@
...
+    }
+  },
+  "v-cache": false,
+  "v-default-fields": [
+    "id",
+    "createdIn",
+    "orderId",
+    "sequenceNumber",
+    "dateSubmitted"
+  ],
```
filafb: Only bring by default the root properties. We can query the other fields when need, using the `_fields` argument. This way we reduce the size of the request back and forth.

masterdata/returnRequest/schema.json
```diff
@@ -0,0 +1,212 @@
...
+  },
+  "v-cache": false,
+  "v-default-fields": [
+    "id",
+    "createdIn",
+    "orderId",
+    "sequenceNumber",
+    "dateSubmitted",
+    "status"
+  ],
```
filafb, line 209: Only bringing by default the root fields and we can query the other fields when needed using the `_fields` params. This way, we reduce the payload in the requests when querying it.

masterdata/returnRequest/schema.json
```diff
@@ -0,0 +1,211 @@
+{
+  "properties": {
+    "orderId": {
+      "type": "string",
+      "required": true
+    },
+    "totalReturnAmount": {
+      "type": "integer",
+      "required": true
+    },
```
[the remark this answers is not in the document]
cesarocampov: I would rename it to totalAvailableReturnAmount

masterdata/returnRequest/schema.json
cesarocampov: what are the possible statuses?

masterdata/returnRequest/schema.json
cesarocampov: @filafb I would add discounts

masterdata/returnRequest/schema.json
cesarocampov: for marketplace seller, the current approach is the best. accounts cannot read external master datas, so a seller won't be able to access the marketplace nor vice versa unless there is a custom service allowing that

masterdata/returnRequest/schema.json
filafb: Those:
``` 
"new",
"processing",
"pickedUpFromClient",
"pendingVerification",
"packageVerified",
"amountRefunded",
"denied"
```

masterdata/returnRequest/schema.json
filafb: I think it falls in the same case as tax. Is it really necessary? I don't see how we would use it down the line. have you thought about any use case?

masterdata/returnRequest/schema.json
```diff
@@ -0,0 +1,212 @@
...
+      "type": "array",
+      "required": true,
+      "items": {
+        "type": "object",
+        "properties": {
+          "orderItemIndex": { "type": "integer", "required": true },
+          "quantity": { "type": "integer", "required": true },
+          "verifiedItems": { "type": ["integer", "null"], "required": true },
+          "reasonCode": { "type": "string", "required": true },
+          "reason": { "type": "string", "required": true },
```
filafb, line 119: Changing the way to handle reason to be more semantic. `otherReason` will be a free text user can enter in the UI. As now, it's saved as `reason`, which can be confusing.

```suggestion
          "returnReason": {
            "type": "object",
            "required": true,
            "properties": {
              "reason": {
                "type": "string",
                "required": true
              },
              "otherReason": {
                "type": "string"
              }
            }
          },
```

masterdata/returnRequest/schema.json
```diff
@@ -0,0 +1,212 @@
...
+        },
+        "iban": {
+          "type": "string"
+        },
+        "accountHolderName": {
+          "type": "string"
+        },
+        "paidWithCc": {
+          "type": "boolean"
+        }
```
filafb, line 106: paidWithCC Isn't necessary here. It should be in the payload when user request the order details.

```suggestion
        }
```

zekivtex (approved):

PR #98, opened 2022-05-05
V3 store styling and products

zekivtex:
Display product image and name, internationalize messages and style them.

How to test it?
[this workspace](https://details--powerplanet.myvtex.com/account?__bindingAddress=www.we-demostore.com/es#/my-returns/add/1229632959691-01)

Quick look:
Captura de Pantalla 2022-05-05 a la(s) 11 39 35


react/store/createReturnRequest/components/ItemsDetails.tsx
```diff
@@ -130,18 +146,27 @@ export const ItemsDetails = (itemToReturn: ItemToReturn) => {
...
         />
-        {/* TODO Intl */}
-        {reasonError && reasonErrorEmptyValue ? 'Reason is required' : null}
+        {reasonError && reasonErrorEmptyValue ? (
+          <CustomMessage
+            status="error"
+            message="store/return-app.return-item-details.dropdown-reason.error"
+          />
+        ) : null}
         {/* TODO user input when other & error */}
```
filafb, line 155: We can delete this TODO. Your implementation for the validation in the other field covers that!

filafb (changes requested):
Nice! I like how it looks like. You just missed a few validation errors (in the address and contact info). Can you apply the same component to them, please?

Also, double check the alignment of the fields for quantity to return, reason and condition in a different browse - safari maybe? I have some style off on Firefox.

Lastly, let's keep in mind that we will have to add css handles to the components here. We will deal with that later.

filafb (approved):

PR #104, opened 2022-05-16
feat: add alert after submitting request

zekivtex:
Display alert after submitting return request.

How to test it?
[this workspace](https://info--powerplanet.myvtex.com/account?__bindingAddress=www.we-demostore.com/es#/my-returns/add/1229632959691-01)

Captura de Pantalla 2022-05-16 a la(s) 15 25 17


react/store/createReturnRequest/components/ConfirmAndSubmit.tsx
```diff
@@ -28,8 +29,10 @@ export const ConfirmAndSubmit = ({ onPageChange, items }: Props) => {
     MutationCreateReturnRequestArgs
   >(CREATE_RETURN_REQUEST)
 
-  // temp state just to show request id on UI
-  const [requestId, setRequestId] = useState('')
+  const { navigate } = useRuntime()
+
+  const [showSubmittedStatus, setShowSubmittedStatus] = useState(false)
```
filafb, line 34: Creating a state to control when a piece of component has to be visible is fine, but we can try to use some different approaches here that might be simpler and easier to maintain.

1. Instead of initiating `confirmationSatus` as `success` (which tells something that is not true for the application since the confirmation is not successfully yet) we can initiate it as an empty string. Then, we can control the visibility of the component using a check for falsy values to `confirmationStatus`.
2. `useMutation` returns a property called `called`. We can combine that with the loading key to control visibility to the component.

react/store/createReturnRequest/components/ConfirmAndSubmit.tsx
```diff
@@ -28,8 +29,10 @@ export const ConfirmAndSubmit = ({ onPageChange, items }: Props) => {
...
+  const [confirmationStatus, setConfirmationStatus] = useState('success')
```
filafb, line 35: We should not initiate a state with a future value (success will only be the case after submitting the useMutation). Also, when we are controlling state with use mutation and we know all possible values for it, we should type the `useState` function.

react/store/createReturnRequest/components/ConfirmAndSubmit.tsx
```diff
@@ -55,8 +58,9 @@ export const ConfirmAndSubmit = ({ onPageChange, items }: Props) => {
         throw new Error('Error creating return request')
       }
 
-      if (data?.createReturnRequest?.returnRequestId) {
-        setRequestId(data.createReturnRequest.returnRequestId)
+      setShowSubmittedStatus(true)
+      if (errors) {
+        setConfirmationStatus('error')
```
filafb, line 63: The error handler here is a little bit confusing and not catching all possibilities. We can review that together.

react/store/createReturnRequest/components/ConfirmPaymentMethods.tsx
```diff
@@ -7,9 +7,6 @@ interface Props {
 }
 
 const messages = defineMessages({
-  bank: {
-    id: 'store/return-app.confirm-payment-methods.refund-method.p-bank',
-  },
   card: {
```
filafb, line 10: I think we need this flag, don't we? Also, we have the same values in another component. We should use the same strategy that we used on Reason and Condition (a centralized message).

filafb (changes requested):
Just adding the comments here so we can go over it together.

filafb (approved):

PR #106, opened 2022-05-20
V3rmavalues

filafb:
This code changes the resolver to create a new request, and calculates and brings the refundable values for it. The values are broken as `items`, `shipping`, and `tax` and every item in the request now has the `sellingPrice` and `tax`. The taxes are either from `item.tax` (from order) or calculated from the TAXHUB value (non-native tax engines). In some cases I found (unileverb2b), the value coming from the TAXHUB for every item is the total for that item (not per unit). The current implementation also follows this logic, so I handled it like that. The `item.tax` is a unit tax.

### How to test it:
This [workspace](https://v3rmavalues--powerplanet.myvtex.com/) has the branch linked. One can create a new request. When checking it's object, the amounts should be present.

### DIFF
--- a/node/resolvers/createReturnRequest.ts
+++ b/node/resolvers/createReturnRequest.ts
@@ -2,8 +2,7 @@ import { UserInputError } from '@vtex/api'
 import type { MutationCreateReturnRequestArgs } from 'vtex.return-app'
 
 import { createItemsToReturn } from '../utils/createItemsToReturn'
-
-// const createRefundableTotals = (itemsToReturn: ReturnRequest['items']): Record<ReturnRequest['']
+import { createRefundableTotals } from '../utils/createRefundableTotals'
 
 export const createReturnRequest = async (
   _: unknown,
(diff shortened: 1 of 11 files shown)


node/resolvers/createReturnRequest.ts
```diff
@@ -56,19 +59,36 @@ export const createReturnRequest = async (
...
+    totals,
   } = order
 
   // Possible bug here: If someone deletes a request, it can lead to a duplicated sequence number.
   // Possible alternative: Save a key value pair in to VBase where key is the orderId and value is either the latest sequence (as number) or an array with all Ids, so we can use the length to calcualate the next seuqence number.
   const sequenceNumber = `${sequence}-${total + 1}`
 
+  const itemsToReturn = createItemsToReturn(items, orderItems)
+
+  // TODO: Apply the flag to return proportional shipping from admin
```
zekivtex, line 72: Thanks for this comment (:

zekivtex (approved):

PR #108, opened 2022-05-25
feat: add proportional shipping value functionality

zekivtex:
What this PR changes:
Add the functionality to calculate a proportional shipping according to the returned items. This is done by multiplying the returned items amount with the total shipping amount, divided the total value of the order minus discounts.

How to test it:
[workspace](https://pickup--powerplanet.myvtex.com/account?__bindingAddress=www.we-demostore.com/es#/my-returns)
Example 1: 
`OrderId = 1234661288883-01`
`requestId = 43fbd375-dc18-11ec-835d-12de88b0135b`
Here we can see that the proportional shipping is calculate it correctly because we returned two items on an order of 5 same items. The result is 2/5, on this case 200 cents is the proportional shipping value for those items.
Captura de Pantalla 2022-05-25 a la(s) 12 50 17
Captura de Pantalla 2022-05-25 a la(s) 12 50 26

Example 2:
`OrderId = 1234651703184-01`
1. `requestId =  b90f62a6-dc10-11ec-835d-02e6b01173ed`
Captura de Pantalla 2022-05-25 a la(s) 13 05 59

2. `requestId = df6ebf56-dc13-11ec-835d-0e09b3a28fc9`
Captura de Pantalla 2022-05-25 a la(s) 13 07 56

3. `requestId = 34bc6536-dc14-11ec-835d-12d813f3a6fb`
Captura de Pantalla 2022-05-25 a la(s) 13 08 46

On this example we made a return request for every product on the order so we can sum the proportional shipping value of each request and that should be equal to the shipping total that in this case is 680.
On the first one the shippingAmount is 20, on the second 327 and the third is 333. 20 + 327 + 333 = 680.

### DIFF
--- a/node/resolvers/createReturnRequest.ts
+++ b/node/resolvers/createReturnRequest.ts
@@ -1,6 +1,7 @@
 import { UserInputError } from '@vtex/api'
 import type { MutationCreateReturnRequestArgs } from 'vtex.return-app'
 
+import { SETTINGS_PATH } from '../utils/constants'
 import { createItemsToReturn } from '../utils/createItemsToReturn'
 import { createRefundableTotals } from '../utils/createRefundableTotals'
 
@@ -10,7 +11,7 @@ export const createReturnRequest = async (
   ctx: Context
 ) => {
   const {
-    clients: { oms, returnRequest: returnRequestClient },
+    clients: { oms, returnRequest: returnRequestClient, appSettings },
     state: { userProfile },
   } = ctx
 
@@ -69,12 +70,12 @@ export const createReturnRequest = async (
 
   const itemsToReturn = createItemsToReturn(items, orderItems)
 
-  // TODO: Apply the flag to return proportional shipping from admin
-  const shippingAmount = totals.find(({ id }) => id === 'Shipping')?.value ?? 0
+  const settings = await appSettings.get(SETTINGS_PATH, true)
 
   const refundableAmountTotals = createRefundableTotals(
     itemsToReturn,
-    shippingAmount
+    totals,
+    settings?.options?.enableProportionalShippingValue as boolean
   )
 
   const refundableAmount = refundableAmountTotals.reduce(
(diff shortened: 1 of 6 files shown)


node/resolvers/createReturnRequest.ts
```diff
@@ -69,12 +70,12 @@ export const createReturnRequest = async (
 
   const itemsToReturn = createItemsToReturn(items, orderItems)
 
-  // TODO: Apply the flag to return proportional shipping from admin
-  const shippingAmount = totals.find(({ id }) => id === 'Shipping')?.value ?? 0
+  const settings = await appSettings.get(SETTINGS_PATH, true)
```
filafb, line 73: We can parallelize this call with the calls to get order and the request. If settings is empty, we should throw an error.

node/resolvers/createReturnRequest.ts
```diff
@@ -69,12 +70,12 @@ export const createReturnRequest = async (
...
 
   const refundableAmountTotals = createRefundableTotals(
     itemsToReturn,
-    shippingAmount
+    totals,
+    settings?.options?.enableProportionalShippingValue as boolean
```
filafb, line 78: This value might not be a `boolean` and can be `undefined`. In this case, it's better to not cast the type.

node/utils/createRefundableTotals.ts
```diff
@@ -22,7 +29,22 @@ export const createRefundableTotals = (
...
+      100
+  }
+
+  const shippingAmount =
+    proportionalShippingAmount !== 0
+      ? {
+          id: 'shipping' as const,
+          value: parseFloat(proportionalShippingAmount.toFixed(2)) * 100,
+        }
+      : { id: 'shipping' as const, value: shippingTotal }
```
filafb, line 47: We can use a single ternary to achieve what we are doing here.

filafb (changes requested):

PR #109, opened 2022-05-26
feat: add loading component to ReturnDetails

zekivtex:
What this PR changes:
Add a custom loader component to `ReturnDetails`.

How to test it:
[workspace](https://skeleton--powerplanet.myvtex.com/account?__bindingAddress=www.we-demostore.com/es#/my-returns/add/1234661288883-01)
loading

PD: I couldnt achieve to make the loader 100% width because on the repo y set as 80%. @filafb do you have a workaround here?

### DIFF
--- a/react/store/createReturnRequest/components/loaders/ReturnDetailsLoader.tsx
+++ b/react/store/createReturnRequest/components/loaders/ReturnDetailsLoader.tsx
@@ -1,0 +1,180 @@
+import React from 'react'
+import type { FunctionComponent } from 'react'
+import { SkeletonPiece, BaseLoading } from 'vtex.my-account-commons'
+import type { OrderToReturnSummary } from 'vtex.return-app'
+import { PageBlock } from 'vtex.styleguide'
+import { FormattedMessage } from 'react-intl'
+
+interface Data {
+  loading: boolean
+  data: OrderToReturnSummary | undefined
+}
+
+interface Props {
+  data: Data
+}
+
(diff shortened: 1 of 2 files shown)


react/store/createReturnRequest/components/loaders/ReturnDetailsLoader.tsx
```diff
@@ -0,0 +1,180 @@
...
+    namespace: 'vtex-account__return-details',
+    title: (
+      <FormattedMessage id="store/return-app.return-order-details.page-header.title" />
+    ),
+    titleId: 'store/return-app.request-return.page.header"',
+    backButton: {
+      titleId: 'store/return-app.return-order-details.page-header.link',
+      path: '#/my-returns/add',
+    },
+  }
```
filafb, line 28: Is there a way to apply the same headerConfig to the ReturnDetails component? I feel they are doing the same thing, but there we are using a PageHeader component. Maybe we can define a object with the `title` component, `linkLabel` and `to` address and use from it. What do you think?

filafb:
> PD: I couldnt achieve to make the loader 100% width because on the repo y set as 80%. @filafb do you have a workaround here?
I think this is fine for now.

react/store/createReturnRequest/components/loaders/ReturnDetailsLoader.tsx
zekivtex: I like the idea of the object!

filafb (approved):

PR #111, opened 2022-05-29
feat: translate select reason, condition and characters left on textarea

zekivtex:
Translate select reason, select condition and characters left messages.

How to test it?
[workspace](https://messages--vtexspain.myvtex.com/account#/my-returns/add)

Captura de Pantalla 2022-05-29 a la(s) 21 08 07


filafb (changes requested):
@zekivtex, can you add the Italian translation so we can show it working? Thanks.

filafb (approved):

PR #112, opened 2022-05-29
Extra settings apart from settings MD document

Git-the-Sanz:
[WORKSPACE](https://sanz--vtexspain.myvtex.com/)

Leveraging the APPS client, a new settings schema was added to the manifest. Allowing customisation of the app without the need of modifying the current masterdata settings schema.

This was included to accomodate a new setting called `displayConditionSelector`, which hides the product condition select from the return request page; by default, this setting is considered `true`, making this change transparent and a minor change.

Screenshot

Screenshot

Screenshot

For minimum impact, I decided to add an extra fetch that resolves the settings and adds them to the original ones, allowing to add more in the future without breaking current implementations.

**A new endpoint and middleware was created for this.**

```js
Promise.all([this.getSettings(), this.getExtraSettings()]).then(([settings, extraSettings]) => {
  if (settings !== null) {
    this.setState({ settings: { ...settings, ...extraSettings }});
  }
})

### DIFF
--- a/react/components/ProductsTable.tsx
+++ b/react/components/ProductsTable.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/restrict-plus-operands */
 import type { FunctionComponent } from 'react'
 import React from 'react'
 import { FormattedCurrency } from 'vtex.format-currency'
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

**Review by zekivtex — APPROVED** [review]

File: react/components/ProductsTable.tsx [hunk 1]
```diff
@@ -1,251 +1,256 @@
-import type { FunctionComponent } from 'react'
```
Git-the-Sanz, line 1: This file was never prettified

react/components/ProductsTable.tsx
```diff
@@ -1,251 +1,256 @@
...
+                    ? formatMessage({
+                        id: `returns.${currentProduct.reasonCode}`,
+                      })
+                    : currentProduct.reasonCode}{' '}
+                  {currentProduct.reasonCode === 'reasonOther'
+                    ? `( ${currentProduct.reason} )`
+                    : null}
+                </div>
+
+                {currentProduct.condition && (
```
Git-the-Sanz, line 121: This was the only change to it

filafb (approved):

PR #117, opened 2022-06-01
fix: display total products value directly

zekivtex:
Fix total amount of products price by calculating directly the sum of each product value instead of relying on the totalPrice from the return request.

How to test it?
Admin
[workspace](https://total--vtexspain.myvtex.com/admin/returns/2a7d42e5-dfe8-11ec-835d-120db15413bb/details/)
Store
[workspace](https://total--vtexspain.myvtex.com/account#/my-returns/details/c7fe72a1-af9b-11ec-835d-123072a6b6a3)

### DIFF
--- a/react/components/ProductsTable.tsx
+++ b/react/components/ProductsTable.tsx
@@ -31,9 +31,7 @@ const ProductsTable: FunctionComponent<Props> = (props) => {
             (parseFloat(currentProduct.tax) || 0)) *
           currentProduct.quantity
 
-        console.log(currentValue, 'current value')
-
-        return (total += currentValue)
+        return total + currentValue
       }, 0)
     : 0
 
--- a/react/components/ProductsTable.tsx
+++ b/react/components/ProductsTable.tsx
@@ -10,7 +10,6 @@ import { renderIcon } from '../common/utils'
 interface Props {
   product: any
   totalRefundAmount: any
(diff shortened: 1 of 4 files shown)


react/components/ProductsTable.tsx
```diff
@@ -22,10 +21,22 @@ const ProductsTable: FunctionComponent<Props> = (props) => {
...
   } = props
 
+  const calculateTotalProductsValue = product.length
+    ? product.reduce((total, currentProduct) => {
+        const currentValue =
+          (currentProduct.unitPrice / 100 +
+            (parseFloat(currentProduct.tax) || 0)) *
+          currentProduct.quantity
+
+        console.log(currentValue, 'current value')
```
filafb, line 34: Remove the logs =)

react/components/ProductsTable.tsx
```diff
@@ -22,10 +21,22 @@ const ProductsTable: FunctionComponent<Props> = (props) => {
...
+
+        return (total += currentValue)
```
filafb, line 36: The result is the same, but I would keep it more simple and do `total + currentValue`. The `+=` mutates the variable `total` and I think we should avoid as most as possible mutating stuff.

filafb (changes requested):

filafb (approved):

PR #118, opened 2022-06-01
V3 dropoff point

ghost:
What this PR changes:

How to test it:
[this workspace](https://dropoff--powerplanet.myvtex.com/admin/graphql-ide).

You can use this query:
```
{
  nearestPickupPoints(
    lat: "2.2015347",
    long: "41.3977365"
  ) {
    items {
      pickupPoint {
        friendlyName
      }
    }
  }
}
```

### DIFF
--- a/node/clients/checkout.ts
+++ b/node/clients/checkout.ts
@@ -1,4 +1,4 @@
-import type { InstanceOptions, IOContext, RequestConfig } from '@vtex/api'
+import type { InstanceOptions, IOContext } from '@vtex/api'
 import { JanusClient } from '@vtex/api'
 import type { NearPickupPointQueryResponse } from 'vtex.return-app'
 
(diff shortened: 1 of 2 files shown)


node/clients/checkout.ts
```diff
@@ -0,0 +1,42 @@
...
+    )
+  }
+
+  protected get = async <T>(url: string, config: RequestConfig = {}) => {
+    try {
+      return await this.http.get<T>(url, config)
+    } catch (e) {
+      return e
+    }
+  }
```
filafb, line 32: This is gonna prevent us to throw errors when they happen. IMO, not a good approach. We can call `this.http.get` inside the methods.

node/typings/pickupPoints.d.ts
```diff
@@ -0,0 +1,56 @@
...
+  id: string
+  additionalInfo: unknown
+  businessHours: Array<{
+    closingTime: string
+    openingTime: string
+    dayOfWeek: number
+  }>
+}
+
+interface CheckoutAddress {
```
filafb, line 30: Can't we leverage the types we created in the GraphQL?

filafb (changes requested):

PR #119, opened 2022-06-01
[Frontend] Replace reasons with custom reasons, when available

Git-the-Sanz:

If the account holder wants to have custom reasons, we will replace the default ones with those.
These custom entries can be translated, so the value/label of the selector will take the user's locale and find if the translation exists, or set the default one as it.

#### How to test it?

The following WS has 2 custom reasons, one without translation and one with a translation for each available locale.
Create a return request and see how one is always the same and the other one reacts to the user's locale.

[Workspace](http://sanz--elefantqa.myvtex.com/)

### DIFF
--- a/react/common/constants/returnsRequest.ts
+++ b/react/common/constants/returnsRequest.ts
@@ -1,0 +1,68 @@
+import { defaultReturnReasonsMessages } from '../../store/utils/defaultReturnReasonsMessages'
+
+export function getReasonOptions(formatMessage) {
+  return [
+    {
+      value: 'reasonAccidentalOrder',
+      label: formatMessage(defaultReturnReasonsMessages.reasonAccidentalOrder),
+    },
+    {
+      value: 'reasonBetterPrice',
+      label: formatMessage(defaultReturnReasonsMessages.reasonBetterPrice),
+    },
+    {
+      value: 'reasonPerformance',
+      label: formatMessage(defaultReturnReasonsMessages.reasonPerformance),
+    },
+    {
+      value: 'reasonIncompatible',
+      label: formatMessage(defaultReturnReasonsMessages.reasonIncompatible),
+    },
+    {
+      value: 'reasonItemDamaged',
+      label: formatMessage(defaultReturnReasonsMessages.reasonItemDamaged),
+    },
+    {
+      value: 'reasonMissedDelivery',
+      label: formatMessage(defaultReturnReasonsMessages.reasonMissedDelivery),
+    },
+    {
+      value: 'reasonMissingParts',
+      label: formatMessage(defaultReturnReasonsMessages.reasonMissingParts),
+    },
+    {
+      value: 'reasonBoxDamaged',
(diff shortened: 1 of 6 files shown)


react/store/createReturnRequest/components/ReturnDetails.tsx
```diff
@@ -77,7 +77,7 @@ export const ReturnDetails = (
           </div>
         </div>
       </div>
-      <ItemsList items={items} />
+      <ItemsList items={items} creationDate={creationDate} />
```
Git-the-Sanz, line 80: ATM we decided to prop drill this value.

react/store/createReturnRequest/components/ItemsDetails.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
```
filafb, line 1:
```suggestion
```

react/common/constants/returnsRequest.ts
```diff
@@ -0,0 +1,68 @@
+import { defaultReturnReasonsMessages } from '../../store/utils/defaultReturnReasonsMessages'
+
+export function getReasonOptions(formatMessage) {
```
filafb, line 3: Can you type `formatMessage`, please? This is fine now because I turned off a flag to allow `any` but I am planning to turn it on when we get rid of the old files.

react/store/createReturnRequest/components/RenderReasonDropdown.tsx
```diff
@@ -35,70 +42,17 @@ export const RenderReasonDropdown = ({
...
-      ),
-    },
-    {
-      value: 'reasonDifferentFromWebsite',
-      label: formatMessage(
-        defaultReturnReasonsMessages.reasonDifferentFromWebsite
-      ),
-    },
-  ]
+  let reasonOptions: Array<{ value: string; label: string }>
```
filafb, line 45: Should we `memo` this variable? The array of custom reasons will be only a few items long (I think), but there is a loop inside a loop. WDYT?

react/store/utils/generateCustomReasonOptions.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+export function generateCustomReasonOptions(
+  customReturnReasons: CustomReturnReason[],
+  locale: string,
+  creationDate: string
+) {
+  return customReturnReasons.reduce(
+    (filteredOptions: Option[], customReason: CustomReturnReason) => {
+      if (!isWithinMaxDaysToReturn(creationDate, customReason.maxDays)) {
```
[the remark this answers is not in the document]
filafb: I will add it into our backlog so we remember it's a potential problem.

filafb (changes requested):

react/common/constants/returnsRequest.ts
Git-the-Sanz: Done

filafb (approved):

Issue #121, opened 2022-06-03
Error getting orders if the email has a `+` sign.


afonso-praca:
When trying to get orders, to initiate a return, if the user has a `+` sign on the email, the orders are not returned. We did an investigation and realized that the email should be encoded when calling the OMS API. So we are asking you to implement that, using the `encodeURIComponent()` function. 

Here go examples of requests for OMS API.
This one works

`http://kevinyeebiz.vtexcommercestable.com.br/api/oms/pvt/orders?clientEmail=kevin.yee@vtex.com&orderBy=creationDate,desc&f_status=invoiced&f_creationDate=creationDate:[2022-05-18T00:00:00.075Z%20TO%202022-06-01T14:22:07.075Z]`

This one does not work

`http://kevinyeebiz.vtexcommercestable.com.br/api/oms/pvt/orders?clientEmail=kevinyee13+buyera@gmail.com&orderBy=creationDate,desc&f_status=invoiced&f_creationDate=creationDate:[2022-05-18T00:00:00.075Z%20TO%202022-06-01T14:22:07.075Z]`

If we do encode the emails, both are going to work.

PR #123, opened 2022-06-07
feat: encode email before get orders

zekivtex:
What this PR changes:
This PR is encoding the email value of the URI to avoid breaking the URL. The issue is mentioned here => https://github.com/vtex-apps/return-app/issues/121

How to test it:
[this workspace](https://v3validation--powerplanet.myvtex.com/account?__bindingAddress=www.we-demostore.com/es#/my-returns/add).

### DIFF
--- a/node/clients/oms.ts
+++ b/node/clients/oms.ts
@@ -25,15 +25,6 @@ type InputInvoiceFields = Omit<
   'invoiceKey' | 'invoiceUrl' | 'courier' | 'trackingNumber' | 'trackingUrl'
 >
 
-interface OrderListParams {
-  clientEmail: string
-  orderBy: 'creationDate,desc'
-  f_status: 'invoiced'
-  f_creationDate: string
-  page: number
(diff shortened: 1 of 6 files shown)


node/clients/oms.ts
```diff
@@ -44,10 +35,9 @@ export class OMSCustom extends OMS {
     })
   }
 
-  public listOrdersWithParams(params?: OrderListParams) {
-    return this.http.get<OrderList>(routes.orders, {
+  public listOrdersWithParams(params?: string) {
+    return this.http.get<OrderList>(`${routes.orders}?${params}`, {
```
filafb, line 39: Any reason to move it from the object to a string? I feel that updating the param as string more difficult than just passing the object to the get request.

zekivtex: It was not working when I encoded the email param. And I didn't find a solution, if you have time we can try together with the params object.

PR #125, opened 2022-06-13
Add contact details component 

spclu6:
What this PR changes:

Add contact details component to admin details


spclu6:
You can test the component in

https://returncontactdetails--powerplanet.myvtex.com/admin/returns/requests/

react/admin/ReturnDetails/components/ContactDetails.tsx
```diff
@@ -0,0 +1,39 @@
...
+  } = data
+
+  return (
+    <section className="flex-ns flex-wrap flex-auto flex-column pt4 pb4">
+      <h3>
+        <FormattedMessage id="admin/return-app.return-request-details.contact-details.title" />
+      </h3>
+      <div className="mb5">
+        <p className="ma0">
+          <FormattedMessage id="returns.name" />: {name}
```
filafb, line 24: Just adding here so we don't forget. Please change the message id to match the pattern we are following in the new version.

filafb (changes requested):

PR #137, opened 2022-06-21
V3 add user comments to timeline

spclu6:

Add the user comments to the return request timeline when the request is created

#### How to test it?

[Workspace](https://returnappusercomments--powerplanet.myvtex.com/admin)

Screenshot 2022-06-21 at 10 24 38

### DIFF
--- a/masterdata/returnRequest/schema.json
+++ b/masterdata/returnRequest/schema.json
@@ -210,6 +210,7 @@
                   "format": "date-time",
                   "required": true
                 },
+                "role": { "type": "string", "required": true },
                 "submittedBy": { "type": "string", "required": true },
                 "visibleForCustomer": { "type": "boolean", "required": true }
               }
--- a/masterdata/returnRequest/schema.json
+++ b/masterdata/returnRequest/schema.json
@@ -208,7 +208,8 @@
                 "createdAt": {
                   "type": "string",
                   "format": "date-time",
-                  "required": true
+                  "required": true,
+                  "enum": ["adminUser", "storeUser"]
                 },
                 "role": { "type": "string", "required": true },
                 "submittedBy": { "type": "string", "required": true },
--- a/masterdata/returnRequest/schema.json
+++ b/masterdata/returnRequest/schema.json
@@ -148,7 +148,6 @@
       }
     },
     "dateSubmitted": { "type": "string", "format": "date-time" },
-    "userComment": { "type": ["string", "null"], "maxLength": 300 },
     "refundData": {
       "type": ["object", "null"],
       "required": true,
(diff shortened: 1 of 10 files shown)


filafb (changes requested):
Nice work! Only I thing. We can remove `userComment` from the master data schema [here](https://github.com/vtex-apps/return-app/blob/124400a40a1b3b31e4264b365e87d3c796cc5513/masterdata/returnRequest/schema.json#L151).

After doing it, please relink the project and run `vtex setup typings`to make sure all types are in place.

filafb:
@cesarocampov what do you think about this UI for user comment? Instead of having it in a different component, we placed it inside the status timeline, creating a single flow for comments, either from store user and admin user.

cesarocampov:
> @cesarocampov what do you think about this UI for user comment? Instead of having it in a different component, we placed it inside the status timeline, creating a single flow for comments, either from store user and admin user.

@filafb @spclu6 I like this approach. 
My only comment would be if we could enable a tag other than the email to identify Customer and Operator responses. 
take a look at the image below:
- the first comment was opened by me on the request. it used my email, this is correct. 
- the second comment was made in the admin of the app as an operator but added the Name of the customer, this is incorrect. 

Screenshot 2022-06-22 at 15 32 55
can we add a string before the comment stating
Customer - email
Operator - email?
what do you think?

filafb:
> the second comment was made in the admin of the app as an operator but added the Name of the customer, this is incorrect.
We get this value from session. Is it possible you were logged in the same browser as store user when you submitted the second comment?

>  can we add a string before the comment stating Customer - email Operator - email? what do you think? 
Let's do it. @spclu6 I will reach out to you on slack to discuss this.

filafb (changes requested):

spclu6:
I added the user role into the return request comments so we can see which user type has made the comment.

Screenshot 2022-06-23 at 12 24 03

masterdata/returnRequest/schema.json
```diff
@@ -211,6 +210,7 @@
                   "format": "date-time",
                   "required": true
                 },
+                "role": { "type": "string", "required": true },
```
filafb, line 213: You can make it as ENUM too. It adds one more layer of check for us.

filafb (changes requested):
Really nice! I love it. @cesarocampov What do you think?

Just asking to make the type on the schema leval a `enum` too.

masterdata/returnRequest/schema.json
cesarocampov: @spclu6 nice!

filafb (approved):

PR #138, opened 2022-06-22
V3 return request list migration

Git-the-Sanz:

The return requests list in V2 was handled as a class component directly consuming masterdata.

Both the table and the filter are independent to the layout they reside on. This will allow us to implement them in the admin and in the store; adding any required difference in their parent wrapper.

For the moment only the `admin` layout implements the list.

Some messages and type/interfaces are not done.

#### How to test it?

[Workspace](https://sanz--powerplanet.myvtex.com/admin/returns/requests/)


react/common/components/returnList/ListTableFilter.tsx
```diff
@@ -0,0 +1,216 @@
...
+  sequenceNumber: '',
+  id: '',
+  createdIn: undefined,
+  orderId: '',
+} as Filters
+
+/**
+ * @todo
+ * - Resolve messages
+ */
```
Git-the-Sanz, line 50: I'm using old V2 message's ids

react/common/components/returnList/StatusActionMenu.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { ActionMenu } from 'vtex.styleguide'
+
+import type { FilterKeys } from './ListTableFilter'
+
+interface Props {
+  handleOnChange: (key: FilterKeys, value: string) => void
+  status: Status | ''
+}
+
+const requestsStatuses = {
```
Git-the-Sanz, line 13: I'm doing the same with this generic variable

react/common/components/returnList/ListTableSchema.tsx
```diff
@@ -0,0 +1,118 @@
...
+        // eslint-disable-next-line react/display-name
+        cellRenderer: ({ cellData }) => {
+          return <div>{renderIcon(cellData)}</div>
+        },
+      },
+    },
+  }
+}
+
+function renderIcon(requestStatus: Status) {
```
Git-the-Sanz, line 59: I didn't check if this was already implemented. Maybe we should have this as 1 only source.

Git-the-Sanz: Lets keep them independent; but what about in the same `utils` file?

react/common/components/returnList/ListTableSchema.tsx
```diff
@@ -0,0 +1,118 @@
...
+  switch (requestStatus) {
+    case status.verified:
+      return (
+        <span className="green">
```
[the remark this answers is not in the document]
Git-the-Sanz: Oh yeah! I missed that, thank you

react/common/components/returnList/ListTableSchema.tsx
```diff
@@ -0,0 +1,118 @@
...
+        title: <FormattedMessage id="returns.sequenceNumber" />,
+        width: 150,
+      },
+      orderId: {
+        title: <FormattedMessage id="returns.orderId" />,
+      },
+      createdIn: {
+        title: <FormattedMessage id="returns.submittedDate" />,
+        cellRenderer: ({ cellData }) => {
+          return new Date(cellData).toLocaleString()
```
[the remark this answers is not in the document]
Git-the-Sanz: True, I forgot that we had this

react/common/components/returnList/JumpToPage.tsx
```diff
@@ -0,0 +1,61 @@
...
+interface Props {
+  handleJumpToPage: (page: number) => void
+  currentPage: number
+  maxPage: number
+}
+
+const JumpToPage = (props: Props) => {
+  const { handleJumpToPage, currentPage, maxPage } = props
+
+  const [desiredPage, setDesiredPage] = useState(0)
```
filafb, line 15: In terms of UX, I think it would be better to initiate the `desiredPage` as the currentPage. This way, if I jump to page 2 then navigate to page one using the previous page button, they values will be aligned. Or every time that current page changes, it goes back to zero, or null. What do you think?

filafb (changes requested):
Great job. The list is really neat. Just add some comments to hear your thoughts and one change that is necessary to avoid a `SyntaxError` in the compiler.

react/common/components/returnList/ListTableSchema.tsx
```diff
@@ -0,0 +1,166 @@
...
+ */
+function renderStatus(requestStatus: Status) {
+  switch (requestStatus) {
+    case status.verified:
+      return (
+        <div className="green flex items-center">
+          <span className="mr2 flex">
+            <IconSuccess size={14} />
+          </span>
+          <FormattedMessage id="admin/return-app-status.package-verified" />
```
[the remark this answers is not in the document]
Git-the-Sanz: I can't as this correlates to the issue with the message prefix :(

react/common/components/returnList/ListTable.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const { returnRequestList } = data ?? {}
+  const { list, paging } = returnRequestList ?? {}
+
+  let pageItemFrom = 0
+  let pageItemTo = 0
+
+  const hasPageAndItems = !!paging && !!list?.length
+
+  if (hasPageAndItems) {
+    const { currentPage, total, perPage, pages } = paging
```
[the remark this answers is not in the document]
Git-the-Sanz: Which error is the TS Server giving you? I don't have anything

react/common/components/returnList/JumpToPage.tsx
Git-the-Sanz: Done

react/common/components/returnList/ListTable.tsx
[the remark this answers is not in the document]
filafb: This:
Screen Shot 2022-07-01 at 4 38 03 PM

Most weirdly is that the compiler is working fine. I replicated the issue in the TS playground [here](https://www.typescriptlang.org/play?ts=3.9.7#code/AQYw9gdgzgLsAOBDA5gSwsgXMA3sR2EArgLYBGApgE7AC+wAPsMQDYuPBEQAmFAZugrdgAXmZE2AKFCRYwABaIoABRQUAgjwCSMCiSijgAQiNI0GSdOCo+wABSKVazdx16oASlxXgM6HDxEOkMzdGQfYCtaSSA) and we do have an error on version 3.9.7 - the same one as in our dependency. On TS v4, it doesn't yell at us [here](https://www.typescriptlang.org/play?ts=4.7.4#code/AQYw9gdgzgLsAOBDA5gSwsgXMA3sR2EArgLYBGApgE7AC+wAPsMQDYuPBEQAmFAZugrdgAXmZE2AKFCRYwABaIoABRQUAgjwCSMCiSijgAQiNI0GSdOCo+wABSKVazdx16oASlxXgM6HDxEOkMzdGQfYCtaSSA).

Maybe the builder is using a more updated version of TS and I am not aware of it?

PR #142, opened 2022-06-28
Feat/confirmation email

edyespinal:


node/resolvers/createReturnRequest.ts
```diff
@@ -169,7 +172,49 @@ export const createReturnRequest = async (
     ],
   })
 
-  // TODO: Send confirmation email.
+  try {
```
filafb, line 175: Nice! I think it's having the try catch here is a nice addition (we usually don't need it inside a resolver). I would only add a comment saying that we are doing it here to avoid sending an error to the browser when only the email fails in the process.

node/clients/mail.ts
```diff
@@ -0,0 +1,38 @@
...
+    return this.http.get(`${TEMPLATE_RENDER_PATH}/${name}`, {
+      metric: 'mail-get-template',
+    })
+  }
+
+  public publishTemplate(template: Template): Promise<any> {
+    return this.http.post(TEMPLATE_RENDER_PATH, template, {
+      headers: {
+        ...this.options?.headers,
+        VtexIdClientAutCookie: this.context.adminUserAuthToken,
```
filafb, line 33: Will we have it when the request come from a live domain and a non VTEX user? If you haven't test it, let's do it!

node/utils/constants.ts
```diff
@@ -9,3 +9,6 @@ export const ORDER_TO_RETURN_VALIDATON: Record<
   OUT_OF_MAX_DAYS: 'OUT_OF_MAX_DAYS',
   ORDER_NOT_INVOICED: 'ORDER_NOT_INVOICED',
 }
+
+export const OMS_RETURN_REQUEST = 'oms-return-request'
```
filafb, line 13: Should we make it more specific, since we will have different templates for different parts of the process?

node/typings/mailClient.d.ts
```diff
@@ -0,0 +1,76 @@
...
+  phoneNumber?: string
+  country?: string
+  locality?: string
+  address?: string
+  paymentMethod?: string
+  iban?: string
+  refundedAmount?: string
+  giftCardCode?: string
+  dateSubmitted?: string
+  [key: string]: string
```
filafb, line 27: This is very generic and our object will take any key and value. Do we want it?

node/utils/templates/omsReturnRequest.ts
```diff
@@ -0,0 +1,40 @@
...
+  ApplicationId: null,
+  IsPersisted: true,
+  IsRemoved: false,
+  Type: '',
+  Templates: {
+    email: {
+      To: '{{data.email}}',
+      CC: null,
+      BCC: '{{#compare data.status "==" \'New\'}}{{/compare}}',
+      Subject: 'Formular de returnare {{data.DocumentId}}',
```
filafb, line 23: Should we translate to English here too?

node/utils/templates/templateMessages.ts
```diff
@@ -0,0 +1,2 @@
...
...                                         <ul class="statusUl statusUlLast">\n                                                {{else}}\n                                                <ul class="statusUl">\n                                                    {{/compare}}\n\n                                                    {{#each comments}}\n                                                    {{#if visibleForCustomer}}\n                                                    <li>{{comment}}</li>\n                                                    {{/if}}\n                                                    {{/each}}\n                                                </ul>\n\n                                                </p>\n                                        </div>\n\n                                        {{/each}}\n                                        {{/if}}\n                                    </div>\n                                </td>\n                            </tr>\n                        </table>\n\n                    </td>\n                </tr>\n\n                {{/compare}}\n\n            </table>\n        </td>\n    </tr>\n</table>\n</body>\n</html>\n'
```
filafb, line 2: This is really hard to read and can led us to errors. For example, somewhere in this template we have a logic like this:
```
{{else compare data.status "==" \'Denied\'}}
```

Since we decided to have individual templates for, maybe we should remove that. To make it easier to we can try to use a html tag like [this one](https://www.npmjs.com/package/html-template-tag). It also might be helpful to create composable parts for the template. Do you want to give it a try?

filafb (changes requested):
I tested and worked fine! Thanks. I added a couple of stuff that I think would help us maintaining it.

PR #146, opened 2022-07-04
Request list visibility changes

Git-the-Sanz:
[WORKSPACE](https://sanzstore--powerplanet.myvtex.com)

---

Fixed: When successfully creating a return request, in My account, the alert CTA to navigate back to the request return list won't trigger a re-render, and the new request will be missing until refreshing the page.

screenshot

---

Added: The return list in my account is now more readable, the Request ID field and filter were removed.

screenshot

---

Added: The admin return list now displays a tooltip that states that the customer now needs to access the detail to retrieve the request ID.

screenshot

---

Fixed: navigation links for both domains (admin/store) now redirect accordingly

---

Fixed: buttons now are displayed as disabled instead of spinning when the request list data is loading

### DIFF
--- a/react/hooks/useReturnRequestList.ts
+++ b/react/hooks/useReturnRequestList.ts
@@ -17,6 +17,7 @@ export const useReturnRequestList = () => {
       page: 1,
     },
     notifyOnNetworkStatusChange: true,
+    fetchPolicy: 'no-cache',
   })
 
   return { returnRequestData: { data, loading, error, refetch } }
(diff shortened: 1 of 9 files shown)


react/hooks/useReturnRequestList.ts
```diff
@@ -17,6 +17,7 @@ export const useReturnRequestList = () => {
       page: 1,
     },
     notifyOnNetworkStatusChange: true,
+    fetchPolicy: 'no-cache',
```
Git-the-Sanz, line 20: This fixes the table not showing the newly created request

filafb (approved):

PR #167, opened 2022-07-27
Fix/v3/locale create endpoint

filafb:

This PR fixes an issue preventing the creation of a RMA via endpoint due to an error on the request translating the items name.

#### How to test it?

To this ENDPOINT https://filafixcreaterma--powerplanet.myvtex.com/_v/return-request, make a POST request to create a new return request.

I considered passing the `locale` directly to `translateItemName` function and then apply it to the method getting the translation. However, the function is called in different places and always in a deeper level then the context where the `locale` was available. Adding it to `context.vtex.locale` in the higher level of the middleware handling the POST to create a request, we ensure locale is available in the context.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+- Add locale into context to allow item names to be translated when creating a new request via API.
+
 ## [3.0.0] - 2022-07-26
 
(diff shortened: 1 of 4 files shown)


Git-the-Sanz (approved):
🚀

cesarocampov (approved):
I do not know if the translation is taking a toll on performance but it took 2.18s to create a new return

filafb:
> I do not know if the translation is taking a toll on performance but it took 2.18s to create a new return

When creating a request, the app calls the order list to validate the items to be returned and this request alone takes ~1.8s to give something back to us.

PR #169, opened 2022-07-27
Feat/invoice rma

filafb:

All the requests made with payment method `sameAsPurchase` will have the transaction automatically refunded - a invoice type Input will be created.

As now, the above behavior only happens when user selects to be refunded via credit card or the payment method was `sameAsPurchase` and the user has used a card to pay for the order. 

The requests are created with `sameAsPurchase` when the store sets `Return Payment Options` to same as Order in the settings for the app - as picture below.

Screen Shot 2022-07-27 at 5 42 40 PM

#### How to test it?

For any request created with `sameAsPurchase`, move the status to `Refunded` in [this workspace](https://filainvoiceallrma--powerplanet.myvtex.com/).

After doing so, there is two ways to check the results:
1. Get the order information with a GET to `https://powerplanet.vtexcommercestable.com.br/api/oms/pvt/orders/:orderId`. On `packageAttachment.packages[number].type` search for the object with type Input. The value and items of that invoice should match the one approved via the App.

2. In the admin side, visit the transaction for that order. One should be able to find a refund with the same value approved in the app.


Git-the-Sanz (approved):
🚀

cesarocampov (approved):
AMAZING!

cesarocampov:
> AMAZING!

@filafb before approving, let's add it as a setting only toggable if the refund method is same as purchase.
Just in case the store has a different method of handling the money

filafb:
@cesarocampov and @Git-the-Sanz I made the feature an option to be selected via admin. The behavior is the same as described as above when the feature is set to be on in the admin page. As other features that read information from the settings, the configuration is saved in the request when it is created. So, if someone changes the admin, and a request was created before it, the request will have the previous configuration.

The main changes:

### Admin Settings changes
- Added a new toggle button in the admin page - react/admin/settings/components/PaymentOptions.tsx. It will be always in the UI, but disabled when admin allows payment selection;
- Added a new field `automaticallyRefundPaymentMethod` in the resolver input and `PaymentOptions` type. graphql/types/ReturnAppSettings.graphql;
- Inside the resolver, the payment method input validation checks if the field `automaticallyRefundPaymentMethod` is present in the request when the field `enablePaymentMethodSelection` is false (admin doesn't allow user to select a payment method). When `enablePaymentMethodSelection` is true, the new field `automaticallyRefundPaymentMethod` is set to be `null`.  node/utils/appSettingsValidation.ts;
- The SettingProvider in the React side ensures that the field `automaticallyRefundPaymentMethod` is present in the request when `enablePaymentMethodSelection` is false. react/admin/settings/provider/SettingsProvider.tsx;

### Create RMA flow
- Added a new field `automaticallyRefundPaymentMethod` in the MD schema. When a new request is created, we validate and use the settings value for the same field to create the return request object. We save it as `null` for cases where the payment method is not `sameAsPurchase`. node/services/createReturnRequestService.ts;
- A new field in the the resolver to get the return request object, so we can use that information in the UI; graphql/types/ReturnRequest.graphql.

### Update RMA flow
- There is a new component in the UI to show admin users if the request will be automatically refunded (a invoice type Input will be created); react/common/components/ReturnDetails/RefundMethodDetail.tsx
- When updating the order to `refunded` status, the resolver will check if the request has `sameAsPurchase` payment method and the field `automaticallyRefundPaymentMethod` set as true to create the invoice. node/utils/handleRefund.ts

Screenshot 2022-08-01 at 15-13-50 https __filainvoiceallrma--powerplanet myvtex com

Git-the-Sanz (approved):
🚀

PR #172, opened 2022-08-02
Request cancellation

Git-the-Sanz:

Previously, return requests couldn't be cancelled. We are adding a way of letting both the admin and store user to cancel a request. This allows the store user to create a new requests with the items from the cancelled request.

#### How to test it?

[Workspace](https://sanzcancellation--powerplanet.myvtex.com/)

Test by creating new return requests as a store user. There're 4 different flows:
- Allowing a cancellation as an admin
- Refuse a cancellation as an admin (Because of the return status)
- Allowing a cancellation as a store user
- Refusing a cancellation as a store user (Because of the return status)

Store user allowed
image

Admin user allowed
image


node/utils/createOrdersToReturnSummary.ts
```diff
@@ -37,7 +37,7 @@ export const createOrdersToReturnSummary = async (
     { page: 1, pageSize: 100 },
     ['items', 'refundData', 'refundPaymentData'],
     undefined,
-    `orderId=${orderId}`
+    `orderId=${orderId} AND status <> cancelled`
```
Git-the-Sanz, line 40: I'm just ignoring any request with their status as cancelled, so the whole resolver won't take them into consideration

react/common/components/ReturnDetails/RequestCancellation.tsx
```diff
@@ -0,0 +1,138 @@
...
+const RequestCancellation = () => {
+  const { isOpen, onOpen, onClose } = utils.useDisclosure()
+  const { data } = useReturnDetails()
+  const {
+    route: { domain },
+  } = useRuntime()
+
+  const { submitting, handleStatusUpdate } = useUpdateRequestStatus()
+
+  if (!data) return null
```
Git-the-Sanz, line 43: Maybe I could replace this with the skeleton loader

node/utils/createOrdersToReturnSummary.ts
filafb: That is a nice trick. I didn't know about that. Super!

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+- New: Ability to cancel a request; this action refunds the items for creating a new return request
```
filafb, line 10:
```suggestion
### Added

Ability to cancel a request; this action allows the store use to create a return request with the same items.
```

node/package.json
```diff
@@ -23,7 +23,6 @@
     "vtex.my-account": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.my-account@1.25.0/public/@types/vtex.my-account",
     "vtex.my-account-commons": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.my-account-commons@1.6.0/public/@types/vtex.my-account-commons",
     "vtex.render-runtime": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.render-runtime@8.132.4/public/@types/vtex.render-runtime",
-    "vtex.return-app": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.return-app@3.0.0-beta.12/public/@types/vtex.return-app",
```
filafb, line 26: Since this changes the types, I think it's worth releasing a new beta, install on your workspace, unlink the project and run `vtex setup --typings` to get the correct package for this types. I know it's kind of annoying, but needed if we want to deploy from github.

Git-the-Sanz:
@filafb I found an issue with this PR, I forgot to move the `Context` files out from /admin/* and in into /common/*
I'll ammend asap

node/services/updateRequestStatusService.ts
```diff
@@ -128,12 +130,22 @@ export const updateRequestStatusService = async (
...
-    throw new NotFoundError('Request not found')
+    throw new NotFoundError(`Request ${requestId} not found`)
+  }
+
+  if (
+    storeUserCancellation &&
+    (returnRequest.status !== 'new' ||
+      returnRequest.customerProfileData.userId !== userId)
+  ) {
+    throw new ForbiddenError(
```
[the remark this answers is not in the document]
filafb: To try to clean it a little bit of the confusing you mentioned I think we can do:

1. Bring the validation below the call to get the request:

```
  const userIsAdmin = Boolean(appkey) || role === 'admin'

  const belongsToStoreUser = returnRequest.customerProfileData.userId === userId

  if (!userIsAdmin && !belongsToStoreUser) {
    throw new ForbiddenError('Not authorized')
  }
```

2. Take the status validation inside the `validateStatusUpdate` - This way we keep everything related to status in the same context. We will need to pass the user info to this, but we keep the concerns separated.

What do you think?

react/common/components/ReturnDetails/RequestCancellation.tsx
filafb: I think it's fine like that.

react/common/components/ReturnDetails/RequestCancellation.tsx
```diff
@@ -0,0 +1,138 @@
...
+        isOpen={isOpen}
+        onClose={onClose}
+        bottomBar={
+          <div className="nowrap">
+            <span className="mr4">
+              <Button
+                size="small"
+                variation="tertiary"
+                onClick={onClose}
+                isLoading={submitting}
```
filafb, line 106: Instead of putting the spinning in the cancel button, I would disable it while submitting. Visually I think it's nice to have only the action we clicked changing its state.

react/store/ReturnRequest/StoreReturnDetails.tsx
```diff
@@ -40,7 +42,14 @@ const StoreReturnDetails = () => {
             to: '#/my-returns',
           })
         }}
-      />
+      >
+        <AlertProvider>
```
filafb, line 46: We don't need to do it right now. Just something for us to think about regarding the structure of the app. Since we are using it now here, we might want to bring it outside the admin folder.

filafb (changes requested):
The flow is super nice and code is neat as always. The only thing that it's not working is the alert for error on an edge case:
1. Store user opens the request details page;
2. Admin user updates the status to processing;
3. Store user tries to cancel the request.

There is an error in the resolver (as expected) but I don't see any error in the browser. It might be because the modal is in front of everything, but it can be confusing for the store user. But since it's a EXTREME edge case, we can don't worry about it now - in that case, I would add it into the backlog.

filafb (approved):

PR #176, opened 2022-08-09
Optional Item Condition

Git-the-Sanz:

The admin now has the ability to require the store user to select, individually, the item condition.
This value is also now visible, if found, in the return details.

#### How to test it?

[Workspace](https://sanzoptionalcondition--powerplanet.myvtex.com/admin)

- Go to return settings and activate/deactivate the last general option setting
- If turned off, you will be able to create a return request without selecting item conditions

Without selector
Screenshot

Condition value now visible for both admin and store user
Screenshot


masterdata/returnRequest/schema.json
```diff
@@ -133,8 +133,9 @@
             "required": ["reason"]
           },
           "condition": {
-            "type": "string",
+            "type": ["string", "null"],
             "enum": [
+              null,
```
Git-the-Sanz, line 138: Due to how schemas behave with enums, `null` is now a valid value for it

react/store/createReturnRequest/CreateReturnRequest.tsx
```diff
@@ -147,10 +147,10 @@ export const CreateReturnRequest = (props: RouteProps) => {
...
-    <OrderToReturnProvider>
-      <StoreSettingsPovider>
+    <StoreSettingsPovider>
+      <OrderToReturnProvider>
         <CreateReturnRequest {...props} />
-      </StoreSettingsPovider>
-    </OrderToReturnProvider>
+      </OrderToReturnProvider>
+    </StoreSettingsPovider>
   )
```
Git-the-Sanz, line 155: Due to `validateNewReturnRequestFields` now requiring the new setting, I had to swap providers

filafb (commented):
Nice, thanks! I feel we need to make a decision about changing the GraphQL schema. I added the options I see for it in the comments.

Also, it's important to keep in mind that if a store has already an admin setup saved and doesn't go to the settings to set it on, the app will behavior as the condition is off. I think it's safe for now, but if someone complains about that we know what is happening.

graphql/types/ReturnRequest.graphql
```diff
@@ -121,7 +121,7 @@ type RefundPaymentData {
 type ReturnRequestItem {
   orderItemIndex: Int!
   quantity: Int!
-  condition: ItemCondition!
+  condition: ItemCondition
```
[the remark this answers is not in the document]
Git-the-Sanz: Regarding the last part, I can refactor the option to be the inverse.
Instead of 'enabling' it, you 'disable' it, so if you don't have it, it's enabled by default.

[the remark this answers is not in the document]
filafb: Since we are still on beta, I think we are fine and it's better to keep the pattern of enabling stuff. It was just for us to keep it in mind.

masterdata/returnRequest/schema.json
Git-the-Sanz: STALE

Git-the-Sanz:
@filafb @zekivtex 
Now, instead of allowing to INPUT `null` and allowing the SCHEMA enum to be `null`, we opted for allowing the INPUT to be nullable and the SCHEMA enum to be 'unspecified'. The TYPE will still be non-nullable, solving the requirement of publishing a new MAJOR version.

filafb (approved):

PR #182, opened 2022-08-16
feat: resolve dateSubmitted in createdIn

zekivtex:

Because of the data migration from v2 to v3 the createdIn value, automatically created by masterdata, will always be the current date and not the date when the return request was created.
Fortunately we have the dateSubmitted field that does respect the creation time.
Solution: resolve dateSubmitted value into createdIn field.

#### How to test it?
[workspace](https://date--vtexspain.myvtex.com/account#/my-returns)

date


node/resolvers/ReturnRequestResponse.ts
```diff
@@ -200,4 +200,10 @@ export const ReturnRequestResponse = {
 
     return cultureInfo
   },
+  // Resolve dateSubmitted value into createdIn field because we lost the original value of createdIn (migration data from v2 to v3).
+  createdIn: async (root: ReturnRequest, _args: unknown) => {
```
filafb, line 204:
```suggestion
  createdIn: async (root: ReturnRequest) => {
```

graphql/types/ReturnRequest.graphql
```diff
@@ -187,6 +187,7 @@ input ReturnRequestFilters {
   sequenceNumber: String
   id: String
   createdIn: DateRangeInput
+  dateSubmitted: DateRangeInput
```
filafb, line 190: Are you adding this field here just in order to use it in the middleware for the list API? If so, we can delete it from here and delete also in the args from the getRequestList.ts. Then, inside the returnRequestListService, we just use the `key === 'createdIn'` to create the filter.

node/middlewares/getRequestList.ts
```diff
@@ -31,6 +36,7 @@ export async function getRequestList(ctx: Context) {
         sequenceNumber: _sequenceNumber as string | undefined,
         id: _id as string | undefined,
         createdIn: _createdIn ? { from, to } : undefined,
+        dateSubmitted: _dateSubmitted ? { from, to } : undefined,
```
filafb, line 39: I think we can remove it.

node/services/returnRequestListService.ts
```diff
@@ -39,10 +39,13 @@ const buildWhereClause = (filter: Maybe<ReturnRequestFilters> | undefined) => {
       return where
     }
 
-    if (key === 'createdIn' && typeof value !== 'string') {
-      where += `${key} between ${filterDate(value.from)} AND ${filterDate(
-        value.to
-      )}`
+    if (
+      (key === 'createdIn' || key === 'dateSubmitted') &&
```
filafb, line 43:
```suggestion
      key === 'createdIn' &&
```

graphql/types/ReturnRequest.graphql
```diff
@@ -187,6 +187,7 @@ input ReturnRequestFilters {
   sequenceNumber: String
   id: String
   createdIn: DateRangeInput
```
filafb, line 189: Let's document it in the code. Please, add a similar comment into the line 73 (createdIn for the type ReturnRequestResponse)

```suggestion
"""
  createdIn: It uses the field dateSubmitted in the return request schema to search for documents.
  The field createdIn is auto generated when the document is created, not reflecting the real value for documents migrated from older versions.
  """
  createdIn: DateRangeInput
```

filafb (changes requested):
Nice! I was thinking about the submittedDate for the API and I think it's better to keep only one, since they will be doing the same thing. It can save some code and I added my idea in the comments. Please let me know what you think. 

Also, update the changelog, please.

filafb (approved):

PR #185, opened 2022-08-17
Responsive store returns

Git-the-Sanz:
### What problem is this solving?

Adjusted all pages for better UI/UX under 639px (phone)

### How to test it?

[Workspace](https://sanzmobile--powerplanet.myvtex.com/)

#### Returns list
- Removed both filters and jump to page
- Page header now has a consistent padding

#### Return detail
- Page header now has a consistent padding
- Table rows now have consistent widths

#### Orders list
- CTAs are now more accesible to the user

#### Create return request
- Table doesn't overflow anymore
- Table actions are now more accesible to the user

#### Confirm and submit
- Card containing summary now has consistent widths
- CTAs are now more accesible to the user

### DIFF
--- a/react/store/createReturnRequest/components/ItemsDetails.tsx
+++ b/react/store/createReturnRequest/components/ItemsDetails.tsx
@@ -3,6 +3,7 @@ import type { ItemCondition } from 'vtex.return-app'
 import { NumericStepper } from 'vtex.styleguide'
 import { useCssHandles } from 'vtex.css-handles'
 import { FormattedMessage } from 'react-intl'
+import { useRuntime } from 'vtex.render-runtime'
 
 import { useReturnRequest } from '../../hooks/useReturnRequest'
 import { CustomMessage } from './layout/CustomMessage'
@@ -38,6 +39,10 @@ export const ItemsDetails = (props: Props) => {
     },
     creationDate,
   } = props
+
+  const {
+    hints: { phone },
+  } = useRuntime()
 
   const handles = useCssHandles(CSS_HANDLES)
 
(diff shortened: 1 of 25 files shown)


react/store/createReturnRequest/components/ItemsList.tsx
```diff
@@ -14,49 +16,100 @@ interface Props {
...
-          </th>
-          <th className="v-mid pv0 tl bb b--muted-4 normal bg-base bt ph3 z1 pv3-s tc">
-            <FormattedMessage id="store/return-app.return-order-details.table-header.reason" />
-          </th>
-          {!enableSelectItemCondition ? null : (
-            <th className="v-mid pv0 tl bb b--muted-4 normal bg-base bt ph3 z1 pv3-s tc">
-              <FormattedMessage id="store/return-app.return-order-details.table-header.condition" />
-            </th>
-          )}
+          {phone ? mobileOrder.map(TableHeader) : desktopOrder.map(TableHeader)}
```
filafb, line 109: I love the pattern here, but I would make it more explicit to avoid [this case of memory leak](https://twitter.com/andrestaltz/status/1473669713490235397) if someone changes the return of `TableHeaderRenderer`

```suggestion
          {phone ? mobileOrder.map((header) => TableHeader(header)) : desktopOrder.map((header) => TableHeader(header))}
```

react/styles.global.css
```diff
@@ -41,4 +41,28 @@ ul.status {
...
+
+.return-itemsList-body tr {
+  border-bottom: 1px solid #e3e4e6;
+}
+
+.return-itemsList-body tr:last-child {
+  border-bottom: none;
+}
+
+.createReturnRequestContainer .vtex-pageHeader__container {
```
filafb, line 54: Nitpicking here. Should we use kebab-case here?

filafb (changes requested):
This is coming really good. Looks even better in the phone. Just added some comments, more related to DX.

And not sure if it's a configuration on my phone, but the style for the disabled fields are impossible to read. Do we have the power to change that or it's in the styleguide component level the problem?

IMG_0990

react/styles.global.css
Git-the-Sanz: Added

react/store/createReturnRequest/components/ItemsList.tsx
Git-the-Sanz: Added

react/store/createReturnRequest/components/ItemsDetails.tsx
```diff
@@ -209,6 +230,18 @@ export const ItemsDetails = (props: Props) => {
...
+        <>
+          <td className={`${handles.detailsTdWrapper} pa4`}>
+            <p className={`${handles.itemsDetailText} tc`}>{quantity}</p>
+          </td>
+          <td className={`${handles.detailsTdWrapper} pa4`}>
+            <p className={`${handles.itemsDetailText} tc`}>
+              {availableToReturn}
+            </p>
+          </td>
+        </>
```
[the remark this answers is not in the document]
Git-the-Sanz: Added decoupled component

Git-the-Sanz:
@filafb looks like the issue comes directly from SAFARI, for some reason the hex color for c-disabled (#979899) is not the same in their engine. Pretty weird if you ask me.

Git-the-Sanz:
@filafb @zekivtex Applied Ezequias' suggested idea, and it works:

image

filafb (approved):

PR #187, opened 2022-08-17
fix: dont return before dispatch

zekivtex:

Allow to unselect all payment methods and if thats the case throw an error indicating that we must select al least one if `sameAsOrder` is on.

#### How to test it?

[Workspace](https://validation--vtexspain.myvtex.com/admin/returns/settings)

Before:
validation_before

After:
after-validation


filafb (changes requested):
The current implementation has a similar effect already, preventing use to unselect all. The main problem we saw - and we are not fixing it with this new approach - is when the account just install the app, clicks to allow to choose, than hit save. There won't be any validation error. To do so, I think we will need to lift the state for the error and the validation function for payment method on PaymentMethod component to RMASettings. Then, when we run `handleSubmit` function, we run the validation to check it. We need to keep the validation for when we try to unselect all too.

CHANGELOG.md
```diff
@@ -7,17 +7,25 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
 - Dynamic messages declared statically
 
 ## [3.4.0] - 2022-08-17
+
 ### Fixed
+
 - Prevent order list and order details paegs to show negative numbers for available items.
 - Avoid having a minus sign when restock fee is 0.
 - English translations.
+- Allow to unselect all payment methods and if thats the case throw an error indicating that we must select al least one if `sameAsOrder` is on.
```
filafb, line 28: Please, rebase and bring the changes under unreleased.

react/admin/settings/RMASettings.tsx
```diff
@@ -54,13 +85,18 @@ export const RMASettings = () => {
...
 
     const maxCustomOptionsDays = customReturnReasons?.reduce(
       (maxDay, option) => (maxDay > option.maxDays ? maxDay : option.maxDays),
       0
     )
 
+    const isAnyPaymentMethod = Object.values(allowedPaymentTypes).some(
+      (paymentMethod) => paymentMethod === true
+    )
+
```
filafb, line 99: The function you brought to this file can validate this and the if statement at once for you.

react/admin/settings/RMASettings.tsx
```diff
@@ -71,9 +107,53 @@ export const RMASettings = () => {
...
+      type: 'updatePaymentOptions',
+      payload: {
+        ...appSettings.paymentOptions,
+        allowedPaymentTypes: updatedPaymentOptions,
+      },
+    })
+
+    if (!validateOptions(paymentOptionsPayload)) {
+      setHasPaymentMethodError(true)
+      window.scrollTo(0, 20)
```
filafb, line 149: This is causing a weird behavior. Every time I unselect all the page jumps up. Maybe you want to add this scroll when the user submits, not when they select the option.

filafb (changes requested):
I think we are close. When I submit without any payment option selected, the error feedback is no on my screen. Maybe you want to bring the trick to scroll the page up. Also, you can focus the input, using the useRef - it would be a more elegant solution, but I am not sure the styleguide component takes the ref as props.

filafb (approved):

PR #189, opened 2022-08-18
feat: add iban frontend validation

zekivtex:

Add frontend IBAN validation

#### How to test it?

Place a random IBAN it should trigger the error.
[Workspace](https://iban--vtexspain.myvtex.com/account#/my-returns/add)

Before(v2):
Captura de Pantalla 2022-08-22 a la(s) 11 31 25
Captura de Pantalla 2022-08-22 a la(s) 11 32 02

_Didn't trigger an error because we where only validating if the country code and iban length was correct. But not validating the IBAN structure with mod97 function. If the IBAN is correct this function should returned_ `1`.

Replit:
Captura de Pantalla 2022-08-22 a la(s) 11 29 29

_Here we can see that the behaviour on v2 was wrong. Because this is not a valid number, the mod97 function returned a number different than_ `1`.

After(current implementation):
Captura de Pantalla 2022-08-22 a la(s) 11 51 57


react/store/createReturnRequest/components/PaymentMethods.tsx
```diff
@@ -175,6 +176,16 @@ export const PaymentMethods = ({ canRefundCard }: Props) => {
                 }
               />
             ) : null}
+            {bankDetailsError &&
+            refundPaymentData.iban &&
+            isValidIBANNumber(refundPaymentData.iban) !== 1 ? (
```
filafb, line 181: Bring the validation `!==1` inside the return of `isValidIBANNumber`. This way, your function will "answer" the user calling it with the right answer -> true or false.

react/store/utils/isValidIBANNumber.ts
```diff
@@ -1,4 +1,4 @@
-function mod97(string: any) {
+function mod97(string) {
```
filafb, line 1: TS the argument, please.

filafb (changes requested):

filafb (approved):

PR #190, opened 2022-08-18
feat: add IBAN validation to backend

zekivtex:

Add Backend IBAN validation

#### How to test it?

Place a random IBAN it should trigger the apollo graphql error.
[Workspace](https://backend--vtexspain.myvtex.com/account#/my-returns/add)

Note: I placed a prefer-early-return flag to avoid esLint warning. Do you think there is a better way? 
```
// eslint-disable-next-line vtex/prefer-early-return
  if (refundPaymentMethod === 'bank') {
    if (!iban) {
      throw new ResolverError('IBAN is required', 400)
    }

    if (!accountHolderName) {
      throw new ResolverError('Account holder name is required', 400)
    }

    if (!isValidIBANNumber(iban)) {
      throw new ResolverError('IBAN is not valid', 400)
    }
  }
```

### DIFF
--- a/node/utils/isValidIBANNumber.ts
+++ b/node/utils/isValidIBANNumber.ts
@@ -1,0 +1,116 @@
+function mod97(string: string) {
+  let checksum = string.slice(0, 2) as string | number
+  let fragment
+
+  for (let offset = 2; offset < string.length; offset += 7) {
+    fragment = `${String(checksum)}${string.substring(offset, offset + 7)}`
+    checksum = parseInt(fragment, 10) % 97
+  }
+
+  return checksum
+}
+
+export function isValidIBANNumber(input: string) {
+  type CodeLengths = {
+    [key: string]: number
+  }
+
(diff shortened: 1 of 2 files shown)


node/utils/isValidIBANNumber.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  const dig = code[3] + code[1] + code[2]
+  const digits = dig.replace(/[A-Z]/g, (letter) => {
+    const returnVar = letter.charCodeAt(0) - 55
+
+    return returnVar.toString()
+  })
+
+  // If the IBAN is correct the result should be 1 if not any number higher than 1.
+  return mod97(digits)
```
filafb, line 116: Same thing as the other PR.
```suggestion
  return mod97(digits) === 1
```

filafb (changes requested):

filafb (approved):

PR #195, opened 2022-08-25
feat: dont rely on sellerName when creating a request

ghost:

Rely only on the sellerId to link a product to a seller. The solution was to get the seller name when resolving the objects that serve it to the front end.

#### How to test it?

[Workspace](https://seller--vtexspain.myvtex.com/admin/graphql-ide)

The requestId provided here is useful for testing because the sellerName was saved as an empty string. If you look for the `sellerName` on the graphqlIDE you should see the correct sellerName
```
{
  returnRequest(requestId: "1eed2aae-2399-11ed-835d-0e0377f22cbd") {
    items {
      sellerName
    }
  }
}

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Add IBAN validation in frontend and backend
 - Remove IBAN and accountHolderNumber if refund method different than bank
+- Stop relying on saving the sellerName at the creation of a return request. Instead get the seller name when resolving the objects that serve it to the front end.
 
 ## [3.4.1] - 2022-08-18
 


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Add IBAN validation in frontend and backend
 - Remove IBAN and accountHolderNumber if refund method different than bank
+- Stop relying on saving the sellerName at the creation of a return request. Instead get the seller name when resolving the objects that serve it to the front end.
```
filafb, line 14: Please rebase and make sure changelog is right. The changes listed above are already released.

graphql/types/ReturnRequest.graphql
```diff
@@ -136,7 +136,7 @@ type ReturnRequestItem {
   imageUrl: String!
   unitMultiplier: Float!
   sellerId: String!
-  sellerName: String!
+  sellerName: String
```
filafb, line 139: Since we are resolving it before sending the response, we can keep it as non-nullable.

It's important to note that when you change something on GraphQL (also routes on node), depending on the changes we get a warning in the terminal like the below one. If we proceed with these change, we would have to release a new major.
Screen Shot 2022-08-30 at 6 21 18 PM

masterdata/returnRequest/schema.json
```diff
@@ -116,7 +116,7 @@
           "imageUrl": { "type": "string" },
           "unitMultiplier": { "type": "number" },
           "sellerId": { "type": "string" },
-          "sellerName": { "type": "string" },
+          "sellerName": { "type": ["string", "null"] },
```
filafb, line 119: We don't need to make it `null` as well. The field is not listed as a required one, so it can be undefined.

node/resolvers/ReturnRequestResponse.ts
```diff
@@ -131,19 +131,28 @@ export const ReturnRequestResponse = {
...
+      clients: { returnRequest: returnRequestClient, oms },
     } = ctx
 
-    const { items: itemsList } = await returnRequestClient.get(id as string, [
-      'items',
-    ])
+    const { orderId, items: itemsList } = await returnRequestClient.get(
+      id as string,
+      ['orderId', 'items']
+    )
```
filafb, line 143: We don't need to make another request to get the request. It is passed into this function as `root`, hence the type it has. Moreover, not sure the order should be the source of truth here. @cesarocampov told me about a endpoint that brings the seller name giving the seller id.

filafb (changes requested):
Sorry for taking this long to review. I added a couple of questions / change requests. I mentioned a endpoint @cesarocampov mentioned to me, but I am not familiar with it. I am suggesting this because I am not sure if the seller name inside the order gets updated when it changes. Not sure how to test it, but we should try to test it. And if we go down the route to use this endpoint, we need to understand what it returns if an account is no longer a seller in a marketplace. It's worth discussing it.

PR #198, opened 2022-09-02
Export returns module

Git-the-Sanz:

The ability to export all or some return requests into a single downloadable file

#### How to test it?

[Powerplanet](https://sanzexport--powerplanet.myvtex.com/admin/returns/requests/)
[YAMAMAY](https://returnv3--yamamay.myvtex.com/admin/returns/requests)

Screenshot 2022-09-02 at 17 11 36


react/admin/ExportReturnRequestsModal/components/ReportStatus.tsx
```diff
@@ -0,0 +1,79 @@
...
+    <>
+      <p className="f4 mt2 mb3">
+        <FormattedMessage id="admin/return-app.export-module.report.status-title" />
+      </p>
+      <div className="flex items-center justify-between">
+        <p className="mv3">
+          <ButtonPlain
+            variation="plain"
+            href={downloadLink ?? null}
+            disabled={inProgress || !downloadLink || staleLink}
```
filafb, line 30: The disable key is not making the button unclickable but also the link is valid for more than 6 hours. 🤷‍♂️

messages/en.json
```diff
@@ -329,5 +329,26 @@
...
+  "return-app.return-request-details.cancellation.modal.storeRefuse": "<p>Sorry, you need to contact the support team to cancel this request due to its current status.</p>",
+  "admin/return-app.export-module.modal-open": "EXPORT RETURNS",
+  "admin/return-app.export-module.modal-title": "Export module",
+  "admin/return-app.export-module.modal-close": "CLOSE",
+  "admin/return-app.export-module.content.first-paragraph": "The Export module is the system responsible for merging all return requests into a single file called <b>report</b>, this can be then sent to an email or downloaded directly from this module",
+  "admin/return-app.export-module.content.second-paragraph": "Keep in mind that download links are not permanent, they expire after <b>6 hours</b> following their availability",
+  "admin/return-app.export-module.content.warning-paragraph": "It is important to remember that depending on the quantity of return requests, the process can take a long time to finish. If you chose to generate a download link, you can leave the module open and it will update when its available for download",
```
filafb, line 338: Maybe we should rephrase it, since I think the user doesn't need to leave the module open. They can close it and the report will still be running in the background, right?

filafb (changes requested):
Awesome job @Git-the-Sanz. Not sure you have discussed that with @cesarocampov, but I think we need to have a solution to bring the `refundData` information. With the current report, the customers cannot have the whole picture if the request is already approved or refunded. Bringing that information can be complex, but I think it's important to discuss it. Maybe we should have a different report for it.

On another note, I remember that we discussed to bring the `refundStatusData` in different columns for each status. I see we have a column for `new`, but it's only bringing `NA`. Lastly, any idea on how to bring the comments?

Git-the-Sanz:
Thanks @filafb! I don't understand what you mean with `refundData`. Let's talk schedule a meet to talk about it!
Regarding `refundStatusData`, looks like I never amended it. I'll do it asap.

node/utils/verifyReportMap.ts
```diff
@@ -0,0 +1,25 @@
...
+import type { ReturnRequestReport } from '../clients/report'
+
+export const verifyReportMap = async (
+  reportClient: ReturnRequestReport
+): Promise<void> => {
+  try {
+    const remoteMap = await reportClient.getMap(localMap.id)
+
+    /* Tests for deep equality, throws error with a custom message if it fails */
+    assert.deepStrictEqual(remoteMap, localMap, 'unequal-maps')
```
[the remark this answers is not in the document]
Git-the-Sanz: If any key or value is changed in our app's local map, the assert will tell us that the map in the account is different; so we land in the catch and fire a `createOrUpdate` method that ensures that the target account will always have the most up to date map.

Now, I think that if a customer wants to have a `custom` report, there is no way of `NOT` following OUR map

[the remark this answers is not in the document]
Git-the-Sanz: No need in updating the ID, but it can be done if we want, and this will start using that one after its creation in the catch()

react/admin/ExportReturnRequestsModal/components/ReportStatus.tsx
Git-the-Sanz: Yeah... I saw that after I already made everything, I was always working with 'new' reports, it seems that the 6 hour rule ONLY applies to reports made from the OMS and not MASTERDATA. I can go about removing all the values regarding the idea of `staleLink`. But also, reports are suddenly deleted by no apparent reason. It seems that this api is extremely rogue hahaha

Git-the-Sanz: Regarding the clickable despite disabled, I saw it and forgot to address it. I thought about adding a class to the container like 'pointer-disable' and via CSS set the pointer as none. That fixes it, its a vtex.styleguide bug

filafb:
> Thanks @filafb! I don't understand what you mean with `refundData`. Let's talk schedule a meet to talk about it! Regarding `refundStatusData`, looks like I never amended it. I'll do it asap.

The refund data is the object that controls the information regarding what will be refunded for the store user. It has the invoice information, the values that will be refunded but also the items that were approved. That is why it can be complicated to bring the info in the same report. But let's chat about it.

And maybe it's better to wait this chat before working on the `refundStatusData`. I am seeing a scenario that might be confusing for the final user.

messages/en.json
Git-the-Sanz: Yeah! It's true that it's difficult to convey the idea of how it works.

The report will run autonomously in the server, but the 'polling' state that continuously updates the frontend with the status ONLY runs when the modal is open. So, there's no way of relying to the user that the export finished.
You've got me thinking, maybe we can shed the context provider outside the modal, this could allow us to couple it with your alert provider and send an alert even if the modal is closed! Exactly like the OMS.

In retrospective I wanted to be stingy on the network calls, so the idea of 'closing' the modal was that you shouldn't have events of something that is not the focus on.

Git-the-Sanz: With this in mind, what do you think of:

If you chose to generate a download link, you can close the module, or even the page, and come back later; or you can leave it open, and the status on the right will continuously update until its available for download"

node/utils/verifyReportMap.ts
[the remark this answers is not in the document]
filafb: Ah! I see it now. We just throw an error when it is different than `unequal-maps`. Thanks.

react/admin/ExportReturnRequestsModal/components/ReportStatus.tsx
filafb: We can keep the logic of the `staleLink` if we can prevent the user to click it when it is stale. The button is an anchor tag that doesn't have the disable feature on HTML. So it's either via CSS as you mention or removing the href prop when it's stale.

messages/en.json
filafb: I wouldn't change the implementation. No need for the pooling call when the modal is not open. I would go about this:

```
The process can take a long time to finish based on the quantity of return requests. Since it runs in the background, you can close this page and check it later. 
```

react/admin/ExportReturnRequestsModal/components/ReportStatus.tsx
Git-the-Sanz: Changed the functionality of the `href` prop. If the link is stale, clicking it will do nothing

messages/en.json
Git-the-Sanz: I've changed the text to reflect this

PR #200, opened 2022-09-20
[Fix] Allow return request for items with undefined refId's

filafb:
The `refId` is a non required field in the catalog, but the schema for the return asks for it. Requests for items missing it were failing.

Request created with an item that doesn't have `refId` (Powerplanet, product: 47211):
Screen Shot 2022-09-20 at 3 12 03 PM

I considered changing the master data schema and the graphql type for the returned items, making it non required and nullable, respectively. However, changing the graphql type from `String!` to `String` requires a new major. The implementation suggested here avoids that.


Git-the-Sanz (approved):
🚀

PR #202, opened 2022-10-07
Fix/return gift

filafb:

This code allows a return request to be created when all items in the order are free and the store has the flag to calculate proportional shipping on. If the order has more than one item, the proportional shipping will be calculated by dividing the shipping total by the number of items.

#### How to test it?

Create an order that has only free items;
In [this workspace](https://filarmafreeproducts--powerplanet.myvtex.com) create a return request.

[This request](https://filarmafreeproducts--powerplanet.myvtex.com/account?__bindingAddress=www.we-demostore.com/es#/my-returns/details/12816e5d-467f-11ed-83ab-02c602764f21) was created for a [order](https://powerplanet.myvtex.com/admin/orders/1267140508706-01) that has only free items.


Git-the-Sanz (approved):
🚀

PR #206, opened 2023-02-02
Fix List Orders permission error

georgebrindeiro:

Recently, the Order Management system for VTEX changed their permissions system. Before it was possible to use the `OMSViewer` policy in the `manifest.json` file to list orders, now you need the `ListOrders` policy for that.

This caused some services in the Returns App to fail with a "Not found message"

#### How to test it?

Open the workspace and see data returned as expected

[Workspace](https://fixlistorders--yamamay.myvtex.com/returns/getOrders/clientEmail=francisco.benitez@vtex.com.br)

##### BEFORE

Screen Shot 2023-02-02 at 13 53 53

##### AFTER

Screen Shot 2023-02-02 at 13 53 59

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -75,6 +75,9 @@
       "name": "OMSViewer"
     },
     {
+      "name": "ListOrders"
+    },
+    {
       "name": "outbound-access",
       "attrs": {
         "host": "api.vtex.com",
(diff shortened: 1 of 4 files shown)


iago1501 (approved):
LGTM, nice catch 🕺🏻

PR #208, opened 2023-02-27
refactor/product-details

Yokaito:

The motivation for this pull request is the display of products in the return request screens and in the view of the requested return. In which on mobile it is very bad for the client to see the information because it is a table with a horizontal scroll, thus hiding some information from not showing the content as a whole.

#### How to test it?

[Workspace](http://kamity--codeby.myvtex.com/)

> Before
antes-solicitação-feita
antes-solicitar

> After
depois-solicitação-feita
depois-solicitar

What I took into account to carry out this task is that the table itself to show the product on mobile would not work. So I made a small mockup of a card demonstrating all the information using [VTEX Styleguide](https://styleguide.vtex.com/). Therefore, we will have a larger vertical scroll, but in which the customer will always know the information of each product in the return request.


beatrizmaselli (commented):
Can you add a changelog and the CSS Handles used at the README file, please?

Yokaito:
@beatrizmaselli done

beatrizmaselli:
We follow this standards for the CHANGELOG:
https://keepachangelog.com/en/1.1.0/

beatrizmaselli (changes requested):
@Yokaito we follow those standards for the CHANGELOG:
https://keepachangelog.com/en/1.1.0/

react/store/createReturnRequest/components/ItemsDetails.tsx
```diff
@@ -44,7 +44,25 @@ export const ItemsDetails = (props: Props) => {
...
+    'availableToReturnWrapper',
+    'availableToReturnKey',
+    'availableToReturnValue',
+    'quantitySelectorWrapper',
+    'reasonWrapper',
+    'conditionWrapper',
+  ] as const
+
   const handles = useCssHandles(CSS_HANDLES)
+  const handlesPhone = useCssHandles(CSS_HANDLES_PHONE)
```
edyespinal, line 65: There is no need to create a separate array for the CSS Handles.
Additionally, I see that some of this classes are already declared and are used in the table. This could create conflicts when working with them.

react/store/createReturnRequest/components/ItemsDetails.tsx
```diff
@@ -153,6 +171,103 @@ export const ItemsDetails = (props: Props) => {
...
+                  />
+                ) : null}
+              </div>
+            )}
+          </div>
+        </Box>
+      </div>
+    )
+  }
+
```
edyespinal, line 270: Since we are returning the Card if the user is using a mobile phone, can we remove the references from the table (lines 161, 190, 233)

react/common/components/ReturnDetails/ItemDetails/ItemDetailsList.tsx
```diff
@@ -84,6 +86,7 @@ const getItemVerificationStatus = (
 
 export const ItemDetailsList = () => {
   const handles = useCssHandles(CSS_HANDLES)
+  const handlesPhone = useCssHandles(CSS_HANDLES_PHONE)
```
edyespinal, line 89: Can we unify the CSS Handles?
There is no need for a separate array.

edyespinal (changes requested):

pedrovasalmeida:
@beatrizmaselli done
@edyespinal done

react/store/createReturnRequest/components/ItemsDetails.tsx
```diff
@@ -24,6 +24,20 @@ const CSS_HANDLES = [
   'productImageWrapper',
   'productImage',
   'itemsDetailText',
+  'cardWrapper',
+  'productImageWrapper',
+  'productImage',
```
edyespinal, line 29: These two classes are duplicated.
If we are intending to have a different use for them, can we rename them, and if not, remove it?

edyespinal (changes requested):

pedrovasalmeida:
@edyespinal done

beatrizmaselli (approved):
LGTM

feraraujofilho:
New mobile layout with Cards added to PR

feraraujofilho (approved):

react/store/createReturnRequest/components/ItemsDetails.tsx
feraraujofilho: Hi @edyespinal  could you please validate it again?

node/clients/mail.ts
```diff
@@ -38,7 +38,7 @@ export class MailClient extends JanusClient {
     return this.http.post(TEMPLATE_RENDER_PATH, template, {
       headers: {
         ...this.options?.headers,
-        VtexIdClientAutCookie: this.context.adminUserAuthToken,
+        VtexIdClientAutCookie: this.context.adminUserAuthToken ?? '',
```
edyespinal, line 41: Why are we passing this as an empty string?

node/clients/oms.ts
```diff
@@ -57,7 +57,7 @@ export class OMSCustom extends OMS {
       invoice,
       {
         headers: {
-          VtexIdClientAutCookie: this.context.adminUserAuthToken,
+          VtexIdClientAutCookie: this.context.adminUserAuthToken ?? '',
```
edyespinal, line 60: Why are we passing this as an empty string?

CHANGELOG.md
```diff
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
 
+
+## [3.5.8] - 2023-03-16
+### Added
+- Create new layout for My Returns List and Request Returns Available List (change from table to Grid Layout with cards)
+
+## [3.5.7] - 2023-03-15
+### Fixed
+- Change structure of solicitation details and new request (from table to card) in mobile devices.
+
```
edyespinal, line 18: We should leave the `### Added` and `### Fixed` sections under `## [Unreleased]` since actions take care of this to keep the CHANGELOG up to date.

edyespinal (changes requested):

node/clients/mail.ts
lucasayb: Simple validation, maybe it's redundant, but it's a good practice to have it. But it is considering always the `adminUserAuthToken` and only if the `adminUserAuthToken` is `null` or `undefined` it will get an empty string.

node/clients/oms.ts
lucasayb: Simple validation, maybe it's redundant, but it's a good practice to have it. But it is considering always the `adminUserAuthToken` and only if the `adminUserAuthToken` is `null` or `undefined` it will get an empty string.

CHANGELOG.md
lucasayb: Didn't catch what you mean.
Like this? https://github.com/vtex-apps/return-app/pull/208/commits/1e5c2c763a6667b3f3144f8f0b74361ea8372755

node/clients/mail.ts
lucasayb: @edyespinal if you prefer, we can remove it, although it doesn't affect the feature itself.

node/clients/oms.ts
lucasayb: @edyespinal if you prefer, we can remove it, although it doesn't affect the feature itself.

CHANGELOG.md
edyespinal: @lucasayb yes, exactly.
Sorry that wasn't clear enough.

But we should also remove both `## [3.5.8] - 2023-03-16` and the `## [3.5.7] - 2023-03-15`. Github actions will create these and put the relevant information under them.

node/clients/mail.ts
edyespinal: @lucasayb, again, sorry I didn't make myself clear 😅
I understand the purpose of the code, I was just wondering if you were getting any errors or having any issues when publishing email templates.

node/clients/oms.ts
edyespinal: Same as above 😅

edyespinal (approved):

PR #209, opened 2023-03-14
Auth layer

Git-the-Sanz:


node/middlewares/auth.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  const { userProfile, appkey: appKeyState } = state
+
+  // Either userProfile or appKeyState must be on state to continue
+  if (!userProfile && !appKeyState) {
+    throw new AuthenticationError('Request failed with status code 401')
+  }
+
+  console.log(state)
```
beatrizmaselli, line 50: @Git-the-Sanz can you remove this console.log?

node/middlewares/auth.ts
```diff
@@ -0,0 +1,53 @@
...
+
+      const { user, userId } = authenticatedUser
+
+      state.userProfile = {
+        userId,
+        email: user,
+        role: isAdmin ? 'admin' : 'store-user',
+      }
+
+      if (isAdmin) {
```
beatrizmaselli, line 29: At this point, shouldn't we also validate the type of access that admin user or key has in case of access without specific roles? Perhaps the televandas/admin owner role is the only one that should have access to perform this type of action.

beatrizmaselli (approved):
LGTM

